PR #2041, opened 2014-06-13
don't check schema revisions in barclamp_mgmt_lib

mapleoin:
crowbar does this on its own in crowbar_framework/lib/schema_migration.rb I
don't see why we're duplicating that here.

Happy to be proven wrong, though. In which case I'll add a comment in the
source explaining the duplication.


tboerger:
What do we want to do with this pull request? It's running quite long and @mapleoin is not working on that anymore.

rsalevsky:
I close this PR now due to there is no progress.

PR #2075, opened 2015-01-09
Add back commit removing rabbitmq credential creation

vuntz:


tboerger:
:+1:

Please create a trello card for tracking and i can do the missing stuff next week when i'm back :)

vuntz:
Dropped the second commit since we have https://github.com/crowbar/crowbar/pull/2078 now

tboerger:
:+1:

vuntz:
We ship that patch already, so merging.

PR #2080, opened 2015-01-27
Move repos in /srv/tftpboot/suse-{11.3,12.0}/repos and create symlinks from SMT automatically

vuntz:
This does two things:
- move the repos to subdir: this makes more sense and clarifies for which distro is which repo
- reduces some setup complexity by creating the symlinks from SMT automatically

Note that this second bit is not complete yet as it depends on knowing some path from SMT...


tboerger:
Makes sense, of course i have not tested that. LGTM

PR #2082, opened 2015-01-30
Check for barclamp (crowbar) versions before install

bkutil:
As crowbar now uses incompatible navigation configs, we prevent incompatible barclamps from installing.

The appropriate version bumps are in the following PRs:

https://github.com/crowbar/barclamp-openstack/pull/79
https://github.com/crowbar/barclamp-tempest/pull/131
https://github.com/crowbar/barclamp-swift/pull/277
https://github.com/crowbar/barclamp-crowbar/pull/1215
https://github.com/crowbar/barclamp-network/pull/310
https://github.com/crowbar/barclamp-cisco-ucs/pull/11

### DIFF
--- a/releases/development/master/extra/barclamp_install.rb
+++ b/releases/development/master/extra/barclamp_install.rb
@@ -109,14 +109,24 @@ candidates.each do |bc|
     puts "Barclamp at #{bc} has no name, skipping"
     next
   end
+  version = barclamp["barclamp"]["version"].to_i rescue 0
   order = 9999
   if barclamp["crowbar"] and barclamp["crowbar"]["order"] and \
     barclamp["crowbar"]["order"].to_i 
     order = barclamp["crowbar"]["order"].to_i
   end
-  barclamps[name] = { :src => bc, :name => name, :order => order, :yaml => barclamp }
+  barclamps[name] = { :src => bc, :name => name, :order => order, :yaml => barclamp, :version => version }
   debug "barclamp[#{name}] = #{barclamps[name].pretty_inspect}"
 end
+
+debug "checking barclamp versions:"
+barclamps.values.sort_by{|v| v[:order]}.each do |bc|
+  if bc[:yaml]["nav"] && bc[:version] < 1
+    warn "Refusing to install #{bc[:name]} barclamp version < 1 due to incompatible navigation."
+    exit -1
+  end
+end
+debug "done"
 
 debug "installing barclamps:"
 barclamps.values.sort_by{|v| v[:order]}.each do |bc|
--- a/releases/development/master/extra/barclamp_install.rb
+++ b/releases/development/master/extra/barclamp_install.rb
@@ -109,12 +109,11 @@ candidates.each do |bc|
     puts "Barclamp at #{bc} has no name, skipping"
     next
   end
-  version = barclamp["barclamp"]["version"].to_i rescue 0
-  order = 9999
-  if barclamp["crowbar"] and barclamp["crowbar"]["order"] and \
-    barclamp["crowbar"]["order"].to_i 
-    order = barclamp["crowbar"]["order"].to_i
-  end
+
+  # We assume the barclamp and crowbar keys exist and their values are hashes.
+  version = (barclamp["barclamp"]["version"] || 0).to_i
+  order   = (barclamp["crowbar"]["order"] || 9999).to_i
+
   barclamps[name] = { :src => bc, :name => name, :order => order, :yaml => barclamp, :version => version }
   debug "barclamp[#{name}] = #{barclamps[name].pretty_inspect}"
 end
--- a/releases/development/master/extra/barclamp_install.rb
+++ b/releases/development/master/extra/barclamp_install.rb
@@ -121,7 +121,7 @@ end
 debug "checking barclamp versions:"
 barclamps.values.sort_by{|v| v[:order]}.each do |bc|
   if bc[:yaml]["nav"] && bc[:version] < 1
-    fatal("Refusing to install #{bc[:name]} barclamp version < 1 due to incompatible navigation.", "logs")
+    fatal("Refusing to install #{bc[:name]} barclamp version < 1 due to incompatible navigation.", nil, -1)
   end
 end
 debug "done"
--- a/releases/development/master/extra/barclamp_install.rb
+++ b/releases/development/master/extra/barclamp_install.rb
@@ -122,8 +122,7 @@ end
 debug "checking barclamp versions:"
 barclamps.values.sort_by{|v| v[:order]}.each do |bc|
   if bc[:yaml]["nav"] && bc[:version] < 1
-    warn "Refusing to install #{bc[:name]} barclamp version < 1 due to incompatible navigation."
-    exit -1
+    fatal("Refusing to install #{bc[:name]} barclamp version < 1 due to incompatible navigation.", "logs")
   end
 end
 debug "done"
--- a/releases/development/master/extra/barclamp_mgmt_lib.rb
+++ b/releases/development/master/extra/barclamp_mgmt_lib.rb
@@ -58,9 +58,13 @@ def debug(msg)
   puts "DEBUG: " + msg if ENV['DEBUG'] === "true"
 end
 
-def fatal(msg, log)
-  puts "ERROR: #{msg}  Aborting; examine #{log} for more info."
-  exit 1
+def fatal(msg, log = nil, exit_code = 1)
+  str  = "ERROR: #{msg}  Aborting; "
+  if log
+    str += "Examine #{log} for more info."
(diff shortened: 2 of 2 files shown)


releases/development/master/extra/barclamp_install.rb
```diff
@@ -35,6 +35,17 @@ def usage()
   exit
 end
 
+def naviational_barclamps
```
vuntz: Typo (missing g)

releases/development/master/extra/barclamp_install.rb
```diff
@@ -35,6 +35,17 @@ def usage()
...
+  [
+    "cisco-ucs"
+    "crowbar",
+    "network",
+    "openstack",
+    "swift",
+    "tempest",
```
vuntz: Hrm, I don't think we should hardcode the list here. We should instead check if there's a nav key in the yml file.

releases/development/master/extra/barclamp_install.rb
```diff
@@ -109,15 +120,28 @@ def rm_tmpdir(tmpdir)
     puts "Barclamp at #{bc} has no name, skipping"
     next
   end
+  if barclamp["barclamp"] and barclamp["barclamp"]["version"] and \
+    barclamp["barclamp"]["version"].to_i
```
vuntz: Why do we need this line?

releases/development/master/extra/barclamp_install.rb
bkutil: By check, do you mean `if bc[:nav] && bc[:version] < 1` ? Wouldn't that miss the barclamps that had navigation before, but not the nav key?

releases/development/master/extra/barclamp_install.rb
bkutil: Its there to be consistent with extracting 'name' key from the crowbar.yml. Should I change this to version = `barclamp["barclamp"]["version"].to_i rescue nil` or similar?

releases/development/master/extra/barclamp_install.rb
vuntz: Navigation was inside the nav key before too. It's just what is under nav that changed.

releases/development/master/extra/barclamp_install.rb
vuntz: I'd do a `rescue 0`.

vuntz: Btw, we need a `else` in case the key is missing, don't we?

bkutil: > Btw, we need a else in case the key is missing, don't we?

Sorry, which key? Missing "barclamp" will be caught by rescue, missing "version" will be nil, converted to 0 by to_i.

vuntz:
Not tried, but LGTM

releases/development/master/extra/barclamp_install.rb
```diff
@@ -109,15 +109,25 @@ def rm_tmpdir(tmpdir)
...
   end
-  barclamps[name] = { :src => bc, :name => name, :order => order, :yaml => barclamp }
+  barclamps[name] = { :src => bc, :name => name, :order => order, :yaml => barclamp, :version => version }
   debug "barclamp[#{name}] = #{barclamps[name].pretty_inspect}"
 end
 
+debug "checking barclamp versions:"
+barclamps.values.sort_by{|v| v[:order]}.each do |bc|
+  if bc[:yaml]["nav"] && bc[:version] < 1
+    debug "Refusing to install #{bc[:name]} barclamp version < 1 due to incompatible navigation."
```
dirkmueller: I"d prefer something higher than a mere debug message here. info, warn or error sounds more appropriate to me.

bkutil:
@dirkmueller I switched to warn, although I'm not 100% sure, as the whole script consistently uses debug...

releases/development/master/extra/barclamp_install.rb
tboerger: No, as you can see at https://github.com/bkutil/crowbar/blob/validate_versions/releases/development/master/extra/barclamp_mgmt_lib.rb#L57 we only got ```debug``` and ```fatal```. I think this have do be redone in a seperate pull request with some kind of refactoring of the CLI library.

releases/development/master/extra/barclamp_install.rb
```diff
@@ -109,15 +109,25 @@ def rm_tmpdir(tmpdir)
...
   end
-  barclamps[name] = { :src => bc, :name => name, :order => order, :yaml => barclamp }
+  barclamps[name] = { :src => bc, :name => name, :order => order, :yaml => barclamp, :version => version }
   debug "barclamp[#{name}] = #{barclamps[name].pretty_inspect}"
 end
 
+debug "checking barclamp versions:"
+barclamps.values.sort_by{|v| v[:order]}.each do |bc|
+  if bc[:yaml]["nav"] && bc[:version] < 1
+    warn "Refusing to install #{bc[:name]} barclamp version < 1 due to incompatible navigation."
```
tboerger: Where is ```warn``` defined? As you can see on my comment for dirk we only implement ```debug``` and ```fatal```.

releases/development/master/extra/barclamp_install.rb
```diff
@@ -109,15 +109,25 @@ def rm_tmpdir(tmpdir)
     puts "Barclamp at #{bc} has no name, skipping"
     next
   end
+  version = barclamp["barclamp"]["version"].to_i rescue 0
   order = 9999
```
tboerger: So to do it the same as for version maybe this?

```ruby
version = barclamp["crowbar"]["order"].to_i rescue 9999
```

releases/development/master/extra/barclamp_install.rb
tboerger: Why not simply ```fatal```? https://github.com/bkutil/crowbar/blob/validate_versions/releases/development/master/extra/barclamp_mgmt_lib.rb#L61

bkutil: warn is a ruby method :) you're right with the fatal...

releases/development/master/extra/barclamp_install.rb
bkutil: Then the result would be different. Imagine `["crowbar"]["order"] = nil`. The old code would return 9999 (it is checked in the condition that ["order"] value is not nil), the new code would return 0, because `nil` gets converted to 0 by `.to_i`. 

For the version, it does not matter, because 0 is a good default.

tboerger: So skip .to_i? :)

bkutil: That would make the order nil, not raise :)

tboerger: I think we don't need the rescue in both cases. IMHO ```order = barclamp["crowbar"]["order"] || 9999``` should be for both cases a good solution :)

dirkmueller:
+1

releases/development/master/extra/barclamp_install.rb
```diff
@@ -109,15 +109,24 @@ def rm_tmpdir(tmpdir)
...
   end
-  barclamps[name] = { :src => bc, :name => name, :order => order, :yaml => barclamp }
+  barclamps[name] = { :src => bc, :name => name, :order => order, :yaml => barclamp, :version => version }
   debug "barclamp[#{name}] = #{barclamps[name].pretty_inspect}"
 end
 
+debug "checking barclamp versions:"
+barclamps.values.sort_by{|v| v[:order]}.each do |bc|
+  if bc[:yaml]["nav"] && bc[:version] < 1
+    fatal("Refusing to install #{bc[:name]} barclamp version < 1 due to incompatible navigation.", "logs")
```
vuntz: Can we enhance `fatal` to make the second argument optional? There are no logs in this case, so the message is kind of wrong.

tboerger: :+1: That makes totally sense. And maybe make an optional other argument for another exit status?

bkutil: I think we're just snowballing more stuff to an existing patch here :)

releases/development/master/extra/barclamp_install.rb
bkutil: That would make the order a string in most cases ;)

bkutil: Ok, I'll stop teasing you :) Let me fix it.

releases/development/master/extra/barclamp_install.rb
tboerger: You are right... That should be part of a much bigger refactoring :)
Take it like it is.

tboerger:
:+1:

PR #2083, opened 2015-01-30
Merge asset manifests on install/uninstall

vuntz:
We obviously need to copy the manifests in a directory on install to
make sure we keep them. Note that on install, if we see the barclamp
doesn't have a manifest, we ensure that there's no old manifest
installed for that barclamp.

### DIFF
--- a/releases/development/master/extra/barclamp_mgmt_lib.rb
+++ b/releases/development/master/extra/barclamp_mgmt_lib.rb
@@ -77,6 +77,7 @@ def bc_install(from_rpm, bc, bc_path, yaml)
     raise "ERROR: could not install barclamp #{bc} because #{yaml["barclamp"]["crowbar_layout"]} is unknown layout."
   end
   generate_navigation
+  generate_assets_manifest
   catalog bc_path
 end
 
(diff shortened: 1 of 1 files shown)


tboerger:
Please use ```Pathname```, it makes multiple file handling much easier like ```join```, or ```read``` without calling seperate ```File```.

tboerger:
```ruby
def manifests_path
  @manifests_path ||= Pathname.new(CROWBAR_PATH).join("barclamps", "manifests")
end
```

vuntz:
@tboerger actually, I'd prefer to avoid this: this makes a totally different style than what exists in the same code path (in `bc_install_layout_1_ap`), which makes it hard to read. I'm happy to switch to `Pathname`, but really, we should switch everything to it in that case. And I'd prefer another PR for this.

tboerger:
I already mentioned that on the other mgmt lib improvement (related to migrations) opened by you. IMHO the Pathname approach makes it much better readable because you can simply call nearby all ```File```, ```FileUtils``` and more methods on the ```Pathname``` object.

vuntz:
@tboerger I don't disagree it's a good move; I disagree here because if you look a few lines earlier, we have the same code already not using `Pathname`. So we should do the move, but switching everything together, to avoid having two similar code paths look totally different.

tboerger:
So maybe we need to solve that later on...

vuntz:
At this point, this has been tested as a package patch for quite some time => merging.

PR #2084, opened 2015-02-03
Set frame_max setting of rabbitmq to 0 during install (bnc#910815)

vuntz:
The versions of bunny prior to 0.9 don't support frames; but chef
requires version < 0.8. So work around this by disabling frames.

As we know that later versions of chef won't require this, also add a
"reminder crash" to revert this when we have a more recent version of
chef.

http://bugzilla.suse.com/show_bug.cgi?id=910815

### DIFF
--- a/releases/development/master/extra/install-chef-suse.sh
+++ b/releases/development/master/extra/install-chef-suse.sh
@@ -714,6 +714,22 @@ fi
 
 echo_summary "Starting required services"
 
+# Add frame_max setting as environment variable, because it's just a workaround
+if [ $(rpm -q --qf "%{version}" rubygem-chef | cut -d . -f 1) -ne 10 ]; then
+    die "Please remove the frame_max setting in rabbitmq."
+fi
+
+if test -f /etc/rabbitmq/rabbitmq-env.conf && \
+   grep -q ^SERVER_START_ARGS= /etc/rabbitmq/rabbitmq-env.conf && \
+   ! grep -q '^SERVER_START_ARGS=.*-rabbit frame_max 0' /etc/rabbitmq/rabbitmq-env.conf; then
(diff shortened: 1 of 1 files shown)


releases/development/master/extra/install-chef-suse.sh
```diff
@@ -714,15 +714,28 @@ fi
...
+  ]
  }
 ].
 EOF
+else
+    # allow unlimited frame size, as the version of bunny required by chef
+    # doesn't know how to deal with frames; and this setting is enforced with
+    # rabbitmq >= 3.0.0 (see
+    # http://www.rabbitmq.com/blog/2012/11/19/breaking-things-with-rabbitmq-3-0/)
+    sed -i "s/%% {frame_max, 131072},/{frame_max, 0},/g" /etc/rabbitmq/rabbitmq.config
```
vuntz: Hrm, it doesn't just work (need to remove the trailing `,`). Would probably be more clever to add this at the end of the rabbit section in the config. Will need some sed magic for that, I guess :-)

tboerger:
LGTM

vuntz:
Hrm, maybe we can create /etc/rabbitmq/rabbitmq-env.config with something like `RABBITMQ_SERVER_START_ARGS="-rabbit frame_max 0"`

That would be cleaner.

nkrinner:
Using rabbitmq-env.conf seems to work when we drop the RABBITMQ-prefix:
``SERVER_START_ARGS="-rabbit frame_max 0"``

Is there a way to run mkphyscloud with this change?

vuntz:
Thanks @nkrinner, I replaced the PR with this (for reference, old code is https://github.com/vuntz/crowbar/commit/ac6e61e162e452c8877d3f0f45aea283686e6e6f)

dirkmueller:
+1, assuming it works.

PR #2087, opened 2015-02-13
install-chef-suse: Really make SLE12 checks optional

vuntz:
/srv/tftpboot/suse-12.0/install is installed by the crowbar package, so
trying to skip the SLE12 checks if that directory is missing doesn't
make sense.

Instead, use the same check we have in barclamp-provisioner
(/srv/tftpboot/suse-12.0/install/boot/x86_64/common).


dirkmueller:
+1

PR #2089, opened 2015-02-19
only require use of GNU screen over ssh

aspiers:
If we're logged in at the console, use of screen should be optional, since there's no risk of losing access when the network gets reconfigured.

Idea stolen from:

https://github.com/crowbar/barclamp-provisioner/pull/388/files#diff-945d060a9723ccccce160d5b86130cc6R95


tboerger:
LGTM

dirkmueller:
fine with me

PR #2090, opened 2015-03-25
Fix path for barclamp_install.rb

nkrinner:
When installing a barclamp not from rpm, but from a file location, a non-
correct path is used resulting in not installed roles etc. This fixes
the path

Note: This will upload all data bags and roles present in the path. Worth fixing this too?

### DIFF
--- a/releases/stoney/master/extra/barclamp_mgmt_lib.rb
+++ b/releases/stoney/master/extra/barclamp_mgmt_lib.rb
@@ -427,7 +427,7 @@ def bc_install_layout_1_chef(from_rpm, bc, bc_path, yaml)
   log = File.join log_path, "#{bc}.log"
   File.open(log, "a") { |f| f.puts("======== Installing #{bc} barclamp -- #{Time.now.strftime('%c')} ========") }
   debug "Capturing chef install logs to #{log}"
-  chef = File.join bc_path, 'chef'
+  chef = File.join '/opt/dell', 'chef'
   cookbooks = File.join chef, 'cookbooks'
   databags = File.join chef, 'data_bags'
   roles = File.join chef, 'roles'


aspiers:
Hmm, sounds like a code smell hardcoding absolute paths in the middle of code like that.  There should be a better way.  Can you clarify why a non-correct path is used in the scenario you describe?  I don't understand the problem yet.

vuntz:
I don't think this is correct: "when not installing from rpm" more or less means "install from a git clone", and that path does exist.

Or am I missing something?

nkrinner:
I cloned a git repo and manually copied it in into the /opt/dell/ tree. Then I ran `/opt/dell/bin/barclamp_install.rb  /opt/dell/barclamps/neutron`. The path I pass is bc_path in the `barclamp_mgmt_lib.rb`. The script appends the other names to the path and expects to find the roles e.g. in ` /opt/dell/barclamps/neutron/chef/roles` and the data bags in `/opt/dell/barclamps/neutron/chef/data_bags`. See lines 431 to 433 in `barclamp_mgmt_lib.rb`

nkrinner:
Is this wrong usage of the script? I don't get it to work when not pointing it to `/opt/dell/barclamps/<barclamp_name>`

But  I agree that there are prettier ways to fix this.

vuntz:
> The script appends the other names to the path and expects to find the roles e.g. in ` /opt/dell/barclamps/neutron/chef/roles` and the data bags in `/opt/dell/barclamps/neutron/chef/data_bags`.

That's correct. So why don't you have these files if you cloned a git repo and copied it to /opt/dell/barclamps? Oh, I read you copied it to /opt/dell?

nkrinner:
Yes, I copied it to /opt/dell.

Also, this works as long as there are no changes to data bags or roles, which might be why it always worked for me that way.

vuntz:
So yeah, you're supposed to simply copy the whole git clone to /opt/dell/barclamps.

nkrinner:
I see. I could have probably discovered that myself too. Anyway, closing this now.

PR #2092, opened 2015-05-05
Allow Storage-only deploys

tserong:
This rebrands the dialog titles and tweaks repo checks to allow for SUSE
Enterprise Storage only deploys, based on the existence of an
SES-specific barclamp.

### DIFF
--- a/releases/development/master/extra/install-chef-suse.sh
+++ b/releases/development/master/extra/install-chef-suse.sh
@@ -541,8 +541,6 @@ test ! -e $cloud_dir -a -d $smt_dir && ln -s $smt_dir $cloud_dir
 # FIXME: repos that we cannot check yet:
 #   Cloud 6 Pool / Updates: non-existing repos
 REPOS_SKIP_CHECKS+=" SUSE-OpenStack-Cloud-SLE11-6-Pool SUSE-OpenStack-Cloud-SLE11-6-Updates SUSE-OpenStack-Cloud-6-Pool SUSE-OpenStack-Cloud-6-Updates"
-#   Storage 1.0 Pool / Updates: non-existing repos
-REPOS_SKIP_CHECKS+=" SUSE-Enterprise-Storage-1.0-Pool SUSE-Enterprise-Storage-1.0-Updates"
 
 # Checks for SLE11 medias
 MEDIA=/srv/tftpboot/suse-11.3/install
--- a/releases/development/master/extra/install-chef-suse.sh
+++ b/releases/development/master/extra/install-chef-suse.sh
@@ -80,8 +80,16 @@ mkdir -p "`dirname "$LOGFILE"`"
 
 run_succeeded=
 
+is_ses () {
+  [ -d /opt/dell/barclamps/suse-enterprise-storage ]
+}
+
 
 DIALOG_TITLE=" SUSE OpenStack Cloud 6 "
+
+if is_ses; then
+  DIALOG_TITLE=" SUSE Enterprise Storage "
+fi
 
 # Infrastructure for nice output/logging
 # --------------------------------------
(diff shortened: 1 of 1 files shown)


tserong:
See also https://github.com/crowbar/barclamp-crowbar/pull/1266 and https://github.com/crowbar/barclamp-provisioner/pull/412

releases/development/master/extra/install-chef-suse.sh
```diff
@@ -797,6 +822,7 @@ fi
 # in cookbook dependencies)
 #
 for i in crowbar deployer dns ipmi logging network ntp provisioner pacemaker \
+         suse-enterprise-storage \
```
vuntz: This will break in Cloud as this barclamp is not available; I guess we need to make the list of barclamps to install depend on the product?

releases/development/master/extra/install-chef-suse.sh
```diff
@@ -572,6 +586,11 @@ check_repo_tag key     11.3 SUSE-Cloud-5-Updates
 check_repo_tag summary 11.3 SLE11-HAE-SP3-Pool     'SUSE Linux Enterprise High Availability Extension 11 SP3' 'false'
 check_repo_tag repo    11.3 SLE11-HAE-SP3-Updates  'updates://zypp-patches.suse.de/autobuild/SLE_HAE/11-SP3/update/x86_64' 'false'
```
vuntz: Why not put this inside the `is_ses` block too? That would avoid having to set `REPOS_SKIP_CHECK`.

releases/development/master/extra/install-chef-suse.sh
tserong: Actually, I don't think it will break, because the other barclamps in that list (database, rabbitmq, openstack, keystone etc.) don't exist on my test system, and yet the install script runs fine. I just end up with a whole bunch of stuff like `/opt/dell/barclamps//database is not a barclamp, ignoring.` in `/var/log/crowbar/install.log`. I'd expect the same on a cloud deploy where the suse-enterprise-storage barclamp is not present.

releases/development/master/extra/install-chef-suse.sh
tserong: Yeah, that's a better idea, I'll do that.

releases/development/master/extra/install-chef-suse.sh
tserong: That said, let me rework it slightly, just in case ;-)

releases/development/master/extra/install-chef-suse.sh
tserong: Actually, @vuntz, does `$REPOS_SKIP_CHECK` still really do much? Can't I stick 'false' as the last arg to `check_repo_tag` and get the same effect?

vuntz: Well, `$REPOS_SKIP_CHECK` is really intended as a temporary solution until the repos become available. So it should generally not be abused.

In this case, since you just don't care about SLES11, skipping the validation tests just seems like the natural thing to do to me. But maybe I'm missing something?

tserong: So, slight tangent: now that the storage repos *are* available, we shouldn't include them in $REPOS_SKIP_CHECKS for either Storage or Cloud?

As for skipping the validation tests being more natural, you're not missing anything, you just identified a cleaner way to implement something than I did originally ;-)

vuntz: > So, slight tangent: now that the storage repos are available, we shouldn't include them in $REPOS_SKIP_CHECKS for either Storage or Cloud?

Yeah. I was actually wondering why they were still listed there :-)

tserong:
Should be a bit better now...

tboerger:
LGTM

tserong:
Don't merge this yet, I still need to test it since rebasing on master

tserong:
Tested SES-only deploy with rebased patches, works fine.

rsalevsky:
+1

rsalevsky:
@tboerger Can you check again?

tboerger:
Still LGTM

PR #2098, opened 2015-05-20
Fix missing $REQUIRE_SLES11 for Storage-only deploys

tserong:

### DIFF
--- a/releases/development/master/extra/install-chef-suse.sh
+++ b/releases/development/master/extra/install-chef-suse.sh
@@ -581,8 +581,8 @@ if ! is_ses; then
       #1558be86e7354d31e71e7c8c2574031a
 fi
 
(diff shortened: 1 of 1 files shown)


tserong:
I can't believe I missed this in the last commit :-/

rsalevsky:
Is there a reason why the sles_11 repos are not needed for SES?

tserong:
If you do an SES-only deploy, with SES 1.0+ you only deploy SLE 12 nodes, which themselves have no need for the SLE 11 repos, so I thought it made sense to remove the hard requirement.

(The admin node itself probably still wants them, for updates to it, though)

rsalevsky:
Ok, thanks for the explanation. :+1:

tboerger:
+1

PR #2100, opened 2015-06-03
Fix cloud repos

rsalevsky:
Depends on https://github.com/crowbar/barclamp-provisioner/pull/423


vuntz:
See also https://github.com/rsalevsky/crowbar/blob/fix_cloud_repos/releases/development/master/extra/install-chef-suse.sh#L377 that you can drop.

dirkmueller:
+1

PR #2101, opened 2015-06-09
fix package names for Cloud5

bmwiedemann:
we have ruby2.1 there

this only broke when trying to install from git


tboerger:
:+1: for this fix, but I'm pretty sure this is not the only path where it needs to be fixed.

rsalevsky:
+1

vuntz:
Is the list of packages up-to-date, actually?

Also, I'm not sure we want to merge stuff in release/tex/master yet as we're not using the branches yet.

tboerger:
Currently it have to be done against master, and not only for the tex path.

dirkmueller:
-1, should be fixed in master instead of release/tex/master

PR #2103, opened 2015-06-16
Relax requirements for nova_floating and public

rhafer:
Required for: https://github.com/crowbar/barclamp-neutron/pull/227

### DIFF
--- a/scripts/network-json-validator
+++ b/scripts/network-json-validator
@@ -184,6 +184,7 @@ class CrowbarNetwork
   attr_reader :broadcast_addr
   attr_reader :subnet_addr_full
   attr_reader :ranges
+  attr_reader :router
   attr_reader :conduit
   attr_reader :vlan
   attr_reader :use_vlan
@@ -417,18 +418,9 @@ def validate_networks databag
     end
   end
 
-  if not @networks['public'].subnet_addr_full.include?(@networks['nova_floating'].subnet_addr_full)
-    raise "'nova_floating' network must be a subnetwork of 'public' network"
-  end
-
-  if @networks['public'].use_vlan != @networks['nova_floating'].use_vlan
-    raise "The 'Use VLAN' setting (JSON attribute: 'use_vlan') of the 'nova_floating' and 'public' networks must be the same."
-  end
-
-  if @networks['public'].use_vlan
-    if @networks['public'].vlan != @networks['nova_floating'].vlan
-      raise "'nova_floating' network must have the same VLAN configuration as 'public' network"
-    end
+  if !@networks['public'].subnet_addr_full.include?(@networks['nova_floating'].subnet_addr_full) &&
+    @networks['nova_floating'].router.nil?
+    raise "'nova_floating' network must have a router defined when not a subnetwork of 'public' network"
   end
 
   if @networks['bmc'].subnet_addr_full != @networks['admin'].subnet_addr_full


mjura:
This looks good +1

mjura:
@rhafer : Could you also prepare PR for tex or should I do this?

releases/development/master/extra/network-json-validator
```diff
@@ -418,16 +419,11 @@ def validate_networks databag
...
-  if @networks['public'].use_vlan != @networks['nova_floating'].use_vlan
-    raise "The 'Use VLAN' setting (JSON attribute: 'use_vlan') of the 'nova_floating' and 'public' networks must be the same."
-  end
-
-  if @networks['public'].use_vlan
-    if @networks['public'].vlan != @networks['nova_floating'].vlan
-      raise "'nova_floating' network must have the same VLAN configuration as 'public' network"
+    if @networks['public'].use_vlan && @networks['nova_floating'].use_vlan &&
+        @networks['public'].vlan == @networks['nova_floating'].vlan
+      raise "When using the same VLAN 'nova_floating' must be a subnetwork of the 'public' network"
```
vuntz: What if the two networks don't use vlan? Is this just fine, or should they be on the same network?

releases/development/master/extra/network-json-validator
```diff
@@ -418,16 +419,11 @@ def validate_networks databag
...
+    elsif @networks['nova_floating'].router.nil?
+      raise "'nova_floating' network must have a router defined when not a subnetwork of 'public' network"
```
mjura: I think `router` attribute we should add as default in `network.json` template for nova_floating.

releases/development/master/extra/network-json-validator
rhafer: @vuntz Good question. Though I don't have an easy answer to that. There is no strict requirement to have them on the same network, especially if they'd use different physical interfaces (which is something we can't really check for in the validator). But that's even true when they're on the same vlan I guess.

Do you want me to implement such a requirement? Or should we get rid of the vlan <-> subnet check all together?

vuntz: @rhafer I'd say that you should implement/drop the checks that you feel are right :-) I guess I wrote the old checks not knowing what I was doing ;-)

releases/development/master/extra/network-json-validator
rhafer: The default configuration (floating a subnet of public) doesn't require a router to be defined in "floating". But yeah, it might make sense to add one anyway. Thanks for the hint

releases/development/master/extra/network-json-validator
mjura: For me we can drop also this check, by CISCO it is called router-on-stick
http://www.firewall.cx/cisco-technical-knowledgebase/cisco-routers/336-cisco-router-8021q-router-stick.html

releases/development/master/extra/network-json-validator
mjura: Yes, I  think that if we expect to have separate `public` and `floating` network we should add this router attribute.

mjura:
I would propose to leave only following condition

```
if not @networks['public'].subnet_addr_full.include?(@networks['nova_floating'].subnet_addr_full) &&
    @networks['nova_floating'].router.nil?
    raise "'nova_floating' network must have a router defined when not a subnetwork of 'public' network"
end
```

dirkmueller:
@rhafer rebase?

rhafer:
Rebased. And also integrated @mjura 's changes.

I am still wondering if we'd need to check the conduits of "public" and "floating" for being equal (-> they endup on the same physical interface) when "public" and "floating" are subnets. I don't think there is a strict requirement for that so I left it out for now.

rsalevsky:
+1

mjura:
Still +1

scripts/network-json-validator
```diff
@@ -417,18 +418,9 @@ def validate_networks databag
...
-
-  if @networks['public'].use_vlan != @networks['nova_floating'].use_vlan
-    raise "The 'Use VLAN' setting (JSON attribute: 'use_vlan') of the 'nova_floating' and 'public' networks must be the same."
-  end
-
-  if @networks['public'].use_vlan
-    if @networks['public'].vlan != @networks['nova_floating'].vlan
-      raise "'nova_floating' network must have the same VLAN configuration as 'public' network"
-    end
+  if not @networks['public'].subnet_addr_full.include?(@networks['nova_floating'].subnet_addr_full) &&
```
mjura: Arrgh, last comment for this code! We have change `not` to `!`

```
if !@networks['public'].subnet_addr_full.include?(@networks['nova_floating'].subnet_addr_full) &&
```

rsalevsky: our style guide prefers `unless`

rhafer: Though ```unless``` does the wrong thing here. Just as ```not```.  We want the ```not``` to just apply to the result of the ```include?``` call. ```!``` will just do that, as it has a higher precedence than ```&&```

rsalevsky:
+1

dirkmueller:
+1

PR #2104, opened 2015-06-24
Update Gemfile

rsalevsky:
- use byebug instead of debugger (openSUSE 13.2 support)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,7 +1,7 @@
 source 'https://rubygems.org'
 
 group :development do
-  gem 'pry-debugger', '>= 0.2'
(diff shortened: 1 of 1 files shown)


aspiers:
What's wrong with the `:github` syntax?

+1 for the byebug change, but I can't accept it yet since you bundled it together with the other change.

rsalevsky:
I thought this was deprecated put seams this information is wrong. I dropped the change.

tboerger:
:+1: as byebug is the debugger of your choice for ruby >= 2.1

aspiers:
+1

PR #2105, opened 2015-06-25
Replaced guard files with a single file

tboerger:
Currently there are still some quirks required like switching the Gemfile url to http instead of https, but I think this is more related to the ruby packaging and versions on our SLE 11 nodes. In general it seems to work fine, feel free to try it!


rsalevsky:
+1 It's not perfect but it works unlike the old Guardfiles but we should optimize it.

tboerger:
I think we can optimize many parts when we simplify the structure of this repository afterwards.

rsalevsky:
Please add line below to the `.gitignore`
```
.guard-*
```

tboerger:
@rsalevsky there is already an gitignore for these files.

rsalevsky:
@tboerger sorry overlooked this

vuntz:
No objection on my side (not giving a real +1 as I'm not using this and can't say whether it's an improvement or not; I'll let the people using guard files decide)

rsalevsky:
I will merge this PR now, as far as I know there is only @tboerger and me who using this.

PR #2106, opened 2015-06-26
Remove unused/unmaintained stuff

tboerger:
To start some cleanup I have created this branch. Of course it's not finished, some things needs to be discussed, documentation needs to be updated and so on. This is just to get feedback early. Everybody who wants to jump in and make improvements, do it!

* [x] Remove distribution specific folders
* [x] Remove unmaintained release notes
* [ ] Remove sledgehammer
* [ ] Update documentation, hints for future packaging on OBS
* [ ] Remove all this README mess
* [x] Remove the script mess in root folder
* [ ] Create rake tasks that can replace the still required functionality of ```dev``` tool
* [ ] Update release branches and switch OBS packaging to them

Did I miss something? Write a comment!


vuntz:
You can remove these directories and files too:
- dev-setup/
- initrd/
- install*
- packaging/

I also feel most scripts in the root directory could be dropped, but this needs to be double-checked.

aspiers:
This looks like it overlaps with https://trello.com/c/9zb0vIcb

aspiers:
(BTW I don't mean that in a bad way!  Just an FYI...)

tboerger:
> This looks like it overlaps with https://trello.com/c/9zb0vIcb

This pull request comes from SUSE-Cloud, so everybody can jump in and extend it ;)

vuntz:
@tboerger: do you know if the stuff under travis-ci/ is still used?

tboerger:
@vuntz the travis stuff seems to be unused

tboerger:
I think sledgehammer-common and .chef can get removed as well

tboerger:
How can we store the available/usable barclamps for the rake tasks? :)

vuntz:
> I think sledgehammer-common and .chef can get removed as well

Actually, sledgehammer-common contains the source for the script that we use in the discovery image, so we should keep it.

vuntz:
> How can we store the available/usable barclamps for the rake tasks? :)

If you tell me which rake tasks, then I can help you :-)

tboerger:
> If you tell me which rake tasks, then I can help you :-)

This tasks doesn't exist yet. I want to integrate rake tasks that provide a small amount of functionality from the dev tool like cloning all required barclamps for development, switching all barclamp branches at once and maybe more things that we need within development :)

tboerger:
> Actually, sledgehammer-common contains the source for the script that we use in the discovery image, so we should keep it.

Where? Everything is in the root archive within OBS/IBS

vuntz:
> > Actually, sledgehammer-common contains the source for the script that we use in the discovery image, so we should keep it.
> 
> Where? Everything is in the root archive within OBS/IBS

Yes, and what's in the root archive should come from here. I know, it's horrible, but that's how it's currently setup.

> Edit: Maybe rename it to something better like discovery?

Happy to do that when we start discussing getting rid of the crowbar-specific terminology ("barclamps", "sledgehammer", etc.).

(btw, this PR right now is only about removing, nothing is moved/renamed/changed; wanted to keep it that way to make review easier: people can simply look at the list of files to see what got dropped without having to care about the diff)

vuntz:
> > If you tell me which rake tasks, then I can help you :-)
> 
> This tasks doesn't exist yet. I want to integrate rake tasks that provide a small amount of functionality from the dev tool like cloning all required barclamps for development, switching all barclamp branches at once and maybe more things that we need within development :)

@tboerger I'd suggest to wait a couple of weeks, until the CSM workshop, before doing that (if that's okay with you), as I'd like to merge all the git repos in only a couple of repos (crowbar-core, crowbar-openstack, crowbar-ceph, etc.). If we do this, then you might not need the same set of rake tasks.

vuntz:
And to answer your original question (since I'm totally fine having rake tasks): if we were to do it now, I'd probably just have one file with a barclamp per line, and I'd store that in each branch you care about.

vuntz:
Remove WIP from PR title.

tboerger:
You can remove the patches stuff from ```releases/development/master/extra/patches``` as well

vuntz:
@tboerger okay, done

tboerger:
I think all the pill and chef server configs from ```releases/development/master/extra/patches``` are useless as well :).
Some cleanup of the gitignore should be fine as well.

Beside that I'm fine with that.

vuntz:
> I think all the pill and chef server configs from releases/development/master/extra/patches are useless as well :).

I think they're used by the non-SUSE install script, unfortunately.

tboerger:
Other distros are currently broken anyway as they have no maintainence for ages now. I think a good path should be to build all on OBS even Debian/Ubuntu debs or Redhat rpms. That way we can get rid of all this custom install stuff.

vuntz:
> Other distros are currently broken anyway as they have no maintainence for ages now. I think a good path should be to build all on OBS even Debian/Ubuntu debs or Redhat rpms. That way we can get rid of all this custom install stuff.

I agree and disagree at the same time ;-) I agree that they've not been maintained lately, but I think there's value in keeping the scripts for now as they are a good indicator of how things should be done for the install process on the other OS. And actually, there are chances that it still works (as the install process didn't change much).

Btw, packaging is unrelated: this is the install script (like install-suse-cloud), not the media build script. But I do also agree that we want to get rid of this in general in the long term -- I think this should become part of the rails app. And I want us to work in that direction :-) As a later step, though.

jsuchome:
Overall, this looks good, and I'd give +1 to it.

My only concern is that we are making this SUSE specific, relying on the stuff we have elsewhere (OBS, mkcloud). What if someone would like to adapt crowbar (and not opencrowbar) for their distro? Even if current solutions were obsoleted, they possibly pointed to some areas that should be covered by any potential user...

rsalevsky:
Also +1 from my side. I think we should merge this branch otherwise it becomes to complex.

vuntz:
> My only concern is that we are making this SUSE specific, relying on the stuff we have elsewhere (OBS, mkcloud). What if someone would like to adapt crowbar (and not opencrowbar) for their distro? Even if current solutions were obsoleted, they possibly pointed to some areas that should be covered by any potential user...

That's a valid concern, and I don't have a good answer to this. My (bad) answer is:

- the build scripts for non-SUSE haven't been used in ages, and don't match what we believe we can maintain, so they won't get fixed; due to the complexity, I don't think anyone else will fix them either.
- our long-term goal is to simplify the installation of crowbar by making it possible to just install the packages to have something working. But we're not there yet. (That being said, it's for that reason that I kept the non-SUSE install scripts, so we could easily check what they're doing later on)
- as a temporary solution, people can take a SUSE-based admin server (known to be working) and fix the support for installing non-SUSE nodes. That's probably the easiest way forward for now.

But clearly, this is not a good-enough answer for me either.

tboerger:
And i think the long term goal is to build all this stuff on OBS for different distros. That's the way how building multidistro packages on the easiest way.

vuntz:
@tboerger yes, agreed.

So let's move on and merge this for now. I think there's some agreement on the direction we want to take for the future, which is good enough for me.

PR #2107, opened 2015-06-29
Create assets folder while installing

rsalevsky:
Create the public asset folder to prevent error while installing the
barclamps for the first time. This case happens only when setting up the
development envoiremt or installing from git.


tboerger:
envoiremt == environment

Small typo in commit message, otherwise I'm fine with that.

rsalevsky:
@tboerger done

MaximilianMeister:
LGTM

PR #2108, opened 2015-07-01
Fix repo creation for Storage-only deploys

tserong:
check_repo_tag takes 5 arguments, but the `check_repo_tag key`
invocation for SUSE-OpenStack-Cloud-SLE11-6-Updates doesn't use the
fourth argument, so $REQUIRE_CLOUD was in the wrong place, which causes
problems for SES only deploys.


releases/development/master/extra/install-chef-suse.sh
```diff
@@ -634,11 +634,15 @@ if [ -z "$CROWBAR_FROM_GIT" ]; then
     fi
 fi
 
-check_or_create_ptf_repository 11.3 Cloud-PTF
-check_or_create_ptf_repository 12.0 Cloud-PTF
-# Currently we only sign the Cloud-PTF repository
-sign_repositories 11.3 Cloud-PTF
-sign_repositories 12.0 Cloud-PTF
+# TODO: do we need a Storage-PTF repo?
```
vuntz: I'd say yes. Or we could actually just name the repo `PTF` (a change that'd be fine in Cloud 6) and have everyone share it :-)

tserong:
OK, this ought to be about right. See also https://github.com/crowbar/barclamp-provisioner/pull/441

PR #2109, opened 2015-07-03
Rename some directories and files

vuntz:
This is to avoid any potential confusion with the names.


tboerger:
What about just ```scripts``` instead of ```install-scripts```? :)

aplanas:
LGTM +1

tboerger:
Maybe you can remove ```gem 'pry-byebug', '>= 3.1.0'``` from the gemfile as well? And restructure the readmes into the docs folder? :)

vuntz:
@tboerger all done

tboerger:
Does the part 

```
COMPONENT LICENSES:

Please see https://github.com/crowbar/crowbar/wiki/Licenses-for-Components
```

really make sense? I would suggest to put a plain Apache-2.0 license into the LICENSE file. It's obvious that some components may got different licenses than the codebase of crowbar itself.

tboerger:
Just the LICENSE thing, beside that I like that!

vuntz:
@tboerger done

rsalevsky:
+1

tboerger:
:+1:

PR #2110, opened 2015-07-08
Install Manila barclamp if available

dirkmueller:


vuntz:
+1 (I'd put it around heat & ceilometer in the list, but that doesn't matter)

rsalevsky:
+1

PR #2111, opened 2015-07-08
Update Guardfile

rsalevsky:
* The folder stucture changed in the last days so we need also
  adapt the Guardfile.
* Set vi:syntax for vim syntax highlighting

### DIFF
--- a/Guardfile
+++ b/Guardfile
@@ -1,3 +1,4 @@
+# vi:syntax=ruby
 #
 # Copyright 2011-2013, Dell
 # Copyright 2013-2015, SUSE Linux GmbH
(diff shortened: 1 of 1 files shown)


Guardfile
```diff
@@ -1,3 +1,4 @@
+#!/usr/bin/env ruby
```
tboerger: But that's wrong, it's not an executable so this should not get a shebang. for enabling syntax highlighting there are other options.

tboerger:
Except the shebang I'm fine with that.

tboerger:
# vi:syntax=ruby

MaximilianMeister:
:+1:

PR #2112, opened 2015-07-08
Add barclamps/.gitignore

rsalevsky:
Create the barclamps folder and ignore everything inside this folder.
This change makes it easier for crowbar developers because they have cloned
all baclamps inside the barclamps folder.

### DIFF
--- a/barclamps/.gitignore
+++ b/barclamps/.gitignore
@@ -1,0 +1,2 @@
+*
+!.gitignore


tboerger:
I would suggest to create the barclamps folder and to create a ```.gitignore``` within that folder with this content:

```
*
!.gitignore
````

That way we are sure that the folder already exists. Or you create the barclamps folder within the rake task.

tboerger:
:+1:

MaximilianMeister:
:+1:

PR #2113, opened 2015-07-13
[WIP] Allow using $bc.yml instead of crowbar.yml for barclamps

vuntz:
As we want to merge barclamps in one common git repo, we need to rename
crowbar.yml files to files with unique names, hence $bc.yml.

### DIFF
--- a/scripts/barclamp_create.rb
+++ b/scripts/barclamp_create.rb
@@ -39,26 +39,28 @@ require "#{require_dir}/barclamp_mgmt_lib.rb"
     path = ARGV[2] || "#{barclamp_dir}"
     target = File.join path, bc
     
-    if !get_crowbar_yml_path(target).nil?
+    unless get_crowbar_yml_path(target).nil?
       puts "Aborting! A barclamp already exists in '#{target}'."
       exit -3
-    elsif ! ( bc =~ /^[a-zA-Z0-9_]*$/ )
-      puts "You must supply a name to create a barclamp"
-      exit -3
-    else
-      puts "Creating barclamp '#{bc}' into '#{path}' as entity '#{org}'."
-      files = []
-      FileUtils.mkdir target
-      clone = Dir.entries(MODEL_SOURCE).find_all { |e| !e.start_with? '.'}
-      clone.each do |item|
-        files += bc_cloner(item, bc, org, MODEL_SOURCE, target, true)
-      end
-      filelist = "#{bc}-filelist.txt"
-      File.open( filelist, 'w' ) do |out|
-        files.each { |line| out.puts line } 
-      end
-      puts "Barclamp #{bc} created in #{target}.  Review #{filelist} for files created."
     end
+
+    if ! ( bc =~ /^[a-zA-Z0-9_]*$/ )
+      puts "You must supply a valid name to create a barclamp"
+      exit -3
+    end
+
+    puts "Creating barclamp '#{bc}' into '#{path}' as entity '#{org}'."
+    files = []
+    FileUtils.mkdir target
+    clone = Dir.entries(MODEL_SOURCE).find_all { |e| !e.start_with? '.'}
+    clone.each do |item|
+      files += bc_cloner(item, bc, org, MODEL_SOURCE, target, true)
+    end
+    filelist = "#{bc}-filelist.txt"
+    File.open( filelist, 'w' ) do |out|
+      files.each { |line| out.puts line }
+    end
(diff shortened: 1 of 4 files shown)


scripts/barclamp_create.rb
```diff
@@ -39,7 +39,7 @@
     path = ARGV[2] || "#{barclamp_dir}"
     target = File.join path, bc
     
-    if File.exist? File.join target, "crowbar.yml"
+    if !get_crowbar_yml_path(target).nil?
```
tboerger: I would prefer ```unless get_crowbar_yml_path(target).nil?```

scripts/barclamp_install.rb
```diff
@@ -75,15 +75,15 @@ def rm_tmpdir(tmpdir)
     system "tar xzf \"#{src}\" -C \"#{tmpdir}\""
     # get the barclamp name, allow the tar ball to be named <barclampname>-barclamp.t*
     target="#{tmpdir}/#{src.split("/")[-1].split(".")[0].chomp("-barclamp")}"
-    if File.exists?(File.join(target,"crowbar.yml"))
+    if !get_crowbar_yml_path(target).nil?
```
tboerger: I would prefer ```unless get_crowbar_yml_path(target).nil?```

scripts/barclamp_install.rb
```diff
@@ -75,15 +75,15 @@ def rm_tmpdir(tmpdir)
...
       candidates << target
     else
       puts "#{src} is not a barclamp tarball, ignoring."
     end
-  when File.exists?(File.join(src,"crowbar.yml"))
+  when !get_crowbar_yml_path(src).nil?
```
tboerger: I would prefer ```unless get_crowbar_yml_path(target).nil?```

scripts/barclamp_install.rb
```diff
@@ -75,15 +75,15 @@ def rm_tmpdir(tmpdir)
...
     # We were handed something that looks like a path to a barclamp
     candidates << File.expand_path(src)
-  when File.exists?(File.join(BARCLAMP_PATH, src,"crowbar.yml"))
+  when !get_crowbar_yml_path(File.join(BARCLAMP_PATH, src))
```
tboerger: I would prefer ```unless get_crowbar_yml_path(File.join(BARCLAMP_PATH, src)).nil?```

scripts/barclamp_install.rb
```diff
@@ -134,8 +134,8 @@ def rm_tmpdir(tmpdir)
       target=File.join(BARCLAMP_PATH, bc[:src].split("/")[-1])
       if File.directory? target
         debug "target directory #{target} exists"
-        if File.exists? "#{target}/crowbar.yml"
-          debug "#{target}/crowbar.yml file exists"
+        if !get_crowbar_yml_path(target).nil?
```
tboerger: I would prefer ```unless get_crowbar_yml_path(target).nil?```

scripts/barclamp_mgmt_lib.rb
```diff
@@ -67,6 +67,20 @@ def fatal(msg, log = nil, exit_code = 1)
...
+    suggested_bc_name = directory.split(File::SEPARATOR)[-1]
+  end
+
+  # Look for both old name (crowbar.yml) and new name ($barclamp.yml)
+  ["#{suggested_by_name}.yml", "crowbar.yml"].each do |basename|
+    path = File.join(directory, basename)
+    return path if File.exists?(path)
+  end
+
+  nil
```
tboerger: If I see that correctly no need for that as it would be nil anyway.

vuntz: You're correct, but I prefer to be explicit about it to make it clear that it's intended. The alternative is to put a comment, but then it's the same amount of lines :-)

scripts/barclamp_install.rb
vuntz: Hrm, that's a `when`, not a `if` :-)

scripts/barclamp_mgmt_lib.rb
tboerger: That will be complained by rubocop 8)

scripts/barclamp_install.rb
tboerger: Sorry, too much comments in the rush :D

tboerger:
:+1:

rsalevsky:
:+1:

PR #2114, opened 2015-07-16
Fix detection of SLES11 SP3 Update channel (bsc#938478)

dirkmueller:
maintenance team just switched to a different tag, which breaks
the check


vuntz:
+1 for when you'll have tested this :-)

rsalevsky:
codewise +1

PR #2115, opened 2015-07-17
Fix detection of SLE11 SP3 Update channels (bsc#938478)

dirkmueller:
maintenance team just switched to a different tag, which breaks
the existing check. Adjust.


dguitarbite:
+1
Took a look the repositories. The new tags seem to work for me.

dirkmueller:
@dguitarbite Already successfully tested the patch  myself as well

dguitarbite:
@dirkmueller Yes, Just mentioned the reason for +1. I think we need another +1 before merging.

dirkmueller:
its a backport, one +1 should be enough :-)

PR #2121, opened 2015-07-23
Relax requirements for nova_floating and public

rhafer:
Technically there is not requirement for nova_floating and public to be on the
same VLAN and for nova_floating to be a subnet of public. This commit reworks
the requirements of the network validator so that allow floating and public to be
on different networks.

(cherry picked from commit 52b491fe646e29926c6b7c6468e59b5464a6971c)


vuntz:
+1

dirkmueller:
+1

PR #2122, opened 2015-07-24
Fix barclamp_install.rb when not passing a full path

vuntz:
The check was just totally wrong.


tboerger:
Hum, I thought this is not required as ```nil?``` can be skipped quite often for conditions, but if tht fixes an issue :+1:

MaximilianMeister:
makes sense, but just wondering how could that have worked before when calling `barclamp_install.rb` with just the bc_name?

vuntz:
@MaximilianMeister I broke the code last week, so it was totally working until I broke it ;-)

PR #2123, opened 2015-07-24
Check if a pattern provides the pattern we search for

dirkmueller:
In case patterns get renamed, we can check for the
compat provides.

### DIFF
--- a/scripts/install-chef-suse.sh
+++ b/scripts/install-chef-suse.sh
@@ -629,7 +629,7 @@ if [ -z "$CROWBAR_FROM_GIT" ]; then
       pattern=patterns-ses-admin
       pattern_short=ses_admin
     fi
-    if ! rpm -q $pattern &> /dev/null; then
+    if ! rpm -q --whatprovides $pattern &> /dev/null; then
         die "$pattern package is not installed; please install with \"zypper in -t pattern $pattern_short\" or \"zypper in $pattern\". Aborting."
     fi
 fi


vuntz:
+1

tboerger:
LGTM

scripts/install-chef-suse.sh
```diff
@@ -629,7 +629,7 @@ if [ -z "$CROWBAR_FROM_GIT" ]; then
       pattern=patterns-ses-admin
       pattern_short=ses_admin
     fi
-    if ! rpm -q $pattern &> /dev/null; then
+    if ! rpm -q --whatprovides $pattern &> /dev/null; then
         die "$pattern package is not installed; please install with \"zypper in -t pattern $pattern_short\" or \"zypper in $pattern\". Aborting."
```
aspiers: Typo?  Shouldn't this be

    die "$pattern pattern is not installed; ...

?

vuntz: No, because the pattern depends on a pattern package, which is what we are checking here. Yes, it's confusing :-)

aspiers: I was already aware that it was a pattern package, but my point was that a pattern package is still a pattern, at least semantically (IIRC there are 3 different types of pattern, so I guess there's no strict definition).  Otherwise the title of this PR would be "Check if a pattern provides the *package* we search for ..."  But anyway, it doesn't matter :)

PR #2125, opened 2015-08-03
[optional, only when SLE12 move completed] SLE12 bits for install-suse-cloud

vuntz:


dirkmueller:
@vuntz do you want to keep sle11 going (see failed gating check) ?

vuntz:
@dirkmueller no, once we do the move to sle12, we're not going back (imho). The only exception would be SLE11 compute nodes, but we can add that bit back later on.

scripts/install-chef-suse.sh
```diff
@@ -556,72 +544,46 @@ test ! -e $cloud_dir -a -d $smt_dir && ln -s $smt_dir $cloud_dir
...
-  check_repo_tag repo    12.0 SLE12-HA-Pool                       'obsproduct://build.suse.de/SUSE:SLE-12:GA/sle-ha/12/POOL/x86_64' 'false'
-  check_repo_tag repo    12.0 SLE12-HA-Updates                    'obsrepository://build.suse.de/SUSE:Updates:SLE-HA:12:x86_64/update' 'false'
-  check_repo_tag repo    12.0 SUSE-Enterprise-Storage-1.0-Pool    'obsproduct://build.suse.de/SUSE:SLE-12:Update:Products:Cloud5/ses/1/POOL/x86_64' $REQUIRE_STORAGE
-  check_repo_tag summary 12.0 SUSE-Enterprise-Storage-1.0-Updates 'SUSE Enterprise Storage 1.0' $REQUIRE_STORAGE
 fi
 
+check_repo_tag repo    12.0 SLES12-Pool                         'obsproduct://build.suse.de/SUSE:SLE-12:GA/SLES/12/POOL/x86_64'
+check_repo_tag repo    12.0 SLES12-Updates                      'obsrepository://build.suse.de/SUSE:Updates:SLE-SERVER:12:x86_64/update'
+check_repo_tag repo    12.0 SUSE-OpenStack-Cloud-6-Pool         'obsproduct://build.suse.de/SUSE:SLE-12:Update:Products:Cloud6/suse-openstack-cloud/6/POOL/x86_64' $REQUIRE_CLOUD
+check_repo_tag summary 12.0 SUSE-OpenStack-Cloud-6-Updates      'SUSE OpenStack Cloud 6' '' $REQUIRE_CLOUD
```
MaximilianMeister: is the empty `''` that you added intentional? don't see any change to the interface of `check_repo_tag`

vuntz: yes, it was simply missing before.

MaximilianMeister:
i haven't spent a lot of time with `install-chef-suse` yet but the changes here make sense to me.

@vuntz so is this still WIP? you said you tested it until running the crowbar app right?

MaximilianMeister:
I retriggered the build with `want_sles12_admin`
https://ci.suse.de/job/openstack-mkcloud/5802/

@dirkmueller you think we could somehow inject custom parameters into the gating?

vuntz:
@MaximilianMeister well, I guess I can remove the WIP bit. It's just that we can't merge until the port to SLE12 is complete (as this PR breaks SLE11).

aspiers:
@vuntz You could simply duplicate it to `install-chef-sles12.sh` and then change `crowbar.spec` so that on SLES12 it uses that instead of `install-chef-suse.sh`.  Since this is a temporary duplication (i.e. `install-chef-suse.sh` would get deleted after a while) I think that's acceptable.

vuntz:
I moved the bits that don't break SLE11 and that we need to https://github.com/crowbar/crowbar/pull/2126.

This PR will be for what we can change once we drop SLE11 support.

toabctl:
There are still some SLE11 checks available in `scripts/install-chef-suse.sh`. Search for `11`.

toabctl:
Seems that only the last commit (https://github.com/vuntz/crowbar/commit/9a32adf0784c3db77b5957c293fa7e3635b8ad93 ) is left. That's the thing I try to fix with https://github.com/crowbar/crowbar/issues/2128 . @vuntz isn't available so let's continue the discussion on https://github.com/crowbar/crowbar/pull/2128

PR #2126, opened 2015-08-04
Fixes for SLE12 on admin node

vuntz:


scripts/install-chef-suse.sh
```diff
@@ -1114,15 +1112,24 @@ for state in "discovering" "discovered" "hardware-installing" \
...
             die "Sanity check for transitioning to $state failed!"
     fi
-    # chef_or_die "Chef run for $state transition failed!"
-    chef-client
+
+    if [ "$state" == "hardware-installing" ]; then
+        # Use crowbar_register mode for claiming the disks, as the OS is
+        # already installed
+        echo '{ "crowbar_wall": { "registering": true } }' > $CROWBAR_TMPDIR/chef-registering.json
+        chef-client --json-attributes $CROWBAR_TMPDIR/chef-registering.json
```
MaximilianMeister:
```ruby
echo '{ "crowbar_wall": { "registering": true } }' | chef-client -j /dev/stdin
```
should work too. then you dont need to write a file

vuntz: True; anyone else on this? (it's mostly a matter of taste; I tend to prefer the current way, but happy to change to your way if others agree it's more elegant)

MaximilianMeister: just wondering why create something if is not really needed?

vuntz: Because it's slightly more readable (to me -- again, a matter of taste, which is why I'm happy to change if someone else agree with you)

aspiers: Why is it more readable to you?  Genuinely curious.

vuntz: Not really sure, tbh. Maybe stdin is not something I expect chef-client to use for this.

@aspiers if you tell me you feel like @MaximilianMeister, I'll update the PR :-)

aspiers: Well, @MaximilianMeister's version is shorter, and doesn't violate DRY by repeating the same file path twice without a temporary variable (the introduction of which would make your version more correct but even longer).  So in the absence of any concrete counter-arguments, yes I think I prefer his version.  Having said that, I'd be inclined to split the line after the `|` symbol.

vuntz: Done.

scripts/install-chef-suse.sh
```diff
@@ -730,13 +730,11 @@ EOF
 fi
 
 chkconfig rabbitmq-server on
-ensure_service_running rabbitmq-server '^Node .+ with Pid [0-9]+: running'
+ensure_service_running rabbitmq-server
 
 chkconfig couchdb on
 ensure_service_running couchdb
 
-chmod o-rwx /etc/chef /etc/chef/{server,solr}.rb
```
MaximilianMeister: @vuntz do you think this caused jenkins to fail?

`WARN: Did not find config file: /etc/chef/solr.rb, using command line options`

toabctl:
Together with https://github.com/crowbar/crowbar/pull/2127 , this patch works for me (locally tested but only the steps './mkcloud cleanup prepare setupadmin prepareinstcrowbar instcrowbar')

MaximilianMeister:
patch worked for me as well

+1

dirkmueller:
+1

PR #2129, opened 2015-08-07
Fix service running detection

dirkmueller:
previously it was grepping the output of the status command
to check whether a service is running, which is unreliable when
a service failed to run- since the error message contains the
string "running". It turns out the proper way of detecting
whether a service is running is by checking for the exit code
instead.


toabctl:
+1

aspiers:
+1

PR #2134, opened 2015-08-18
when there is no local yml file, filter out zipped nil value

MaximilianMeister:


tboerger:
Wtf? This was missing in my pull request, I have already fixed that... Please squash the commits.

ovisan:
:+1: lgtm

PR #2136, opened 2015-08-19
[WIP]Fix install-chef-suse.sh after crowbar merge

rsalevsky:
Since the crowbar merge we can't use the rpm based file list anymore
due to the rpm's conatins now more then one barclamp.

### DIFF
--- a/scripts/barclamp_mgmt_lib.rb
+++ b/scripts/barclamp_mgmt_lib.rb
@@ -533,7 +533,7 @@ def bc_install_layout_1_chef(from_rpm, bc, bc_path, yaml)
   log = File.join log_path, "#{bc}.log"
   File.open(log, "a") { |f| f.puts("======== Installing #{bc} barclamp -- #{Time.now.strftime('%c')} ========") }
   debug "Capturing chef install logs to #{log}"
-  chef = File.join bc_path, 'chef'
+  chef = File.join BASE_PATH, 'chef'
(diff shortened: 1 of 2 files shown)


scripts/install-chef-suse.sh
```diff
@@ -59,7 +59,7 @@ if [ -z "$HOME" ]; then
     exit 1
 fi
 
-BARCLAMP_INSTALL_OPTS="--rpm"
```
tboerger: Why have you removed that?

scripts/barclamp_mgmt_lib.rb
```diff
@@ -533,7 +533,7 @@ def bc_install_layout_1_chef(from_rpm, bc, bc_path, yaml)
   log = File.join log_path, "#{bc}.log"
   File.open(log, "a") { |f| f.puts("======== Installing #{bc} barclamp -- #{Time.now.strftime('%c')} ========") }
   debug "Capturing chef install logs to #{log}"
-  chef = File.join bc_path, 'chef'
+  chef = File.join BASE_PATH, 'chef'
```
tboerger: I'm not sure if BASE_PATH is correct

rsalevsky:
Please don't merge! This patch is not a final solution but unblocks `install-chef-suse.sh`.

rsalevsky:
Superseded by #2146

PR #2137, opened 2015-08-21
add help and general configuration option

MaximilianMeister:


tboerger:
IMHO :-1:, now you are getting in a direction where rake simply feels wrong.

MaximilianMeister:
@tboerger why? i just added one help task, and and one option.
we will probably need improvements/additions in the dev setup over time, and right now someone who doesn't know the internals would not know how to use it.

@rsalevsky @tboerger shall we extract some logic into a lib instead?

dirkmueller:
I hope this is no longer relevant, if not please speak up.

PR #2139, opened 2015-08-25
Remove spurious argument when checking Cloud 6 update repo

tserong:
Commit db16734 introduced an additional '' (empty string) argument to
the `check_repo_tag summary 12.0 SUSE-OpenStack-Cloud-6-Updates` line,
which breaks SES-only deploys by causing the update repo to always be
created even if it doesn't exist.


rsalevsky:
+1

tboerger:
:+1:

PR #2140, opened 2015-08-27
Set `sudo: false` to run travis tests in docker

rsalevsky:
This change should shorten the time until the travis test start.


aspiers:
Instead of keeping `sudo` enabled for this, can't we make it install `bashate` without root?

PR #2141, opened 2015-08-27
Set `sudo: false` to run travis tests in docker

rsalevsky:
This change should shorten the time until the travis test starts.

See http://docs.travis-ci.com/user/migrating-from-legacy/"

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
-language: bash
-sudo: true
+language: python
+sudo: false
 install:
-    - make install
+  - "pip install bashate"
 script:
-    - make test
+  - make test


rsalevsky:
@aspiers Can you check this solution? We need to set `language: python` otherwise pip did not work.

aspiers:
Please don't remove the `Makefile`, that's needed for local testing.  Why not just keep `make test` in `.travis.yml`?

tboerger:
A make file even if we have a rakefile? Really?

aspiers:
Thanks @rsalevsky for updating this - looks much better now! 

> A make file even if we have a rakefile? Really?

Oh, I forgot about `Rakefile`.  Sure, that makes more sense than `Makefile`.  But the point of my comment was not to be religious about using a `Makefile`; it was that we must have a convenient way of running the tests locally.  Currently that is done via `make` not `rake`.  So I guess you are implying that we should move the `Makefile` tasks to `Rakefile`?  I would be totally fine with that.

tboerger:
Instead of replacing the make task i would prefer a check within the makefile if we are in the travis environment which drops the sudo in front of the pip install.

tboerger:
@aspiers yes, use the rakefile as we already have it in all barclamps and this repository.

PR #2142, opened 2015-08-31
Fix Guard watch for scripts

rsalevsky:
`.+` means one or more characters but we have only the scripts
folder without subfolders


tboerger:
:+1:

MaximilianMeister:
:+1:

PR #2143, opened 2015-08-31
Moving the Makefile logic into Rakefile and setting sudo to false to run travis tests in docker

ovisan:
https://github.com/crowbar/crowbar/pull/2141 can be closed after this PR


tboerger:
Still a python dependency for a ruby project but :+1:

Rakefile
```diff
@@ -223,3 +223,40 @@ namespace :crowbar do
     end
   end
 end
+
+namespace :test do
+  task dependencies: [] do
+    if ENV["TRAVIS"]
+      system("pip install bashate")
```
dirkmueller: perhaps simply do that in the .travis.yml instead?

Rakefile
```diff
@@ -223,3 +223,40 @@ namespace :crowbar do
...
+    else
+      system("sudo pip install bashate")
+    end
+    exit $?.exitstatus
+  end
+
+  desc "Run bashate tests"
+  task bashate: [:dependencies] do
+    script = "scripts/install-chef-suse.sh"
```
dirkmueller: would be nice to keep it as an extensible for loop like the code before did in case there will be additional scripts to check in the future

Rakefile
tboerger: @aspiers wanted to preserve the make tasks... Now they are rake tasks...

Rakefile
tboerger: As this is anyway already a variable name it can be extended more than easy. But a loop for a single file is some kind of weird...

ovisan: But i can't loop over 1 file, usually a loop is used for more items, like the elements of an array.

Rakefile
aspiers: Actually I wanted to preserve the ability to run the tasks locally.

rsalevsky:
LGTM, is nothing critical and can improved later.

PR #2144, opened 2015-09-01
Remove creation of the barclamps data_bag

rsalevsky:
This PR is a test if we can remove the creation of the barclamps data_bag. I found no line where we use it any longer.


tboerger:
I don't know any place as well. Let's see what gating says. For the meantime it looks like a :+1: from my side.

rsalevsky:
Gating is successful, will drop this part in my PR which fixes `barclamp_install.rb`.

tboerger:
But why not simply merging that as this pr is already created?

PR #2146, opened 2015-09-03
Adapt the barclamp installation to handle components

rsalevsky:
Please note that the gating will fail because we have some workarounds in the packages which needs to be removed after the merge.

### DIFF
--- a/scripts/barclamp_install.rb
+++ b/scripts/barclamp_install.rb
@@ -54,20 +54,9 @@ end
 
 usage if ARGV.length < 1
 
-tmpdir = "/tmp/component_install-#{Process.pid}-#{Kernel.rand(65535)}"
-debug "tarball tmpdir: #{tmpdir}"
-Dir.mkdir(tmpdir)
-
 log_path = File.join '/var', 'log', 'crowbar'
 FileUtils.mkdir log_path unless File.directory? log_path
 log = File.join log_path, "component_install.log"
-
-def rm_tmpdir(tmpdir)
-  if File.directory?(tmpdir)
-    debug "temporary directory #{tmpdir} will be removed"
-    system "rm -rf #{tmpdir}"
-  end
-end
 
 barclamp_yml_files = Array.new
 component_paths = Array.new
@@ -139,8 +128,6 @@ barclamps.values.sort_by{|v| v[:order]}.each do |bc|
           debug "crowbar YAML file does not exists in #{target}"
           puts "#{target} exists, but it is not a barclamp."
           puts "Cowardly refusing to overwrite it."
-          debug "temporary directory #{tmpdir} will be removed if it exists"
-          system "rm -rf #{tmpdir}" if File.directory?(tmpdir)
           exit -1
         else
           debug "crowbar YAML file exists in #{target}"
@@ -155,8 +142,6 @@ barclamps.values.sort_by{|v| v[:order]}.each do |bc|
               puts "  cd \"#{target}\"; find -type f -not -name sha1sums -print0 | \\"
               puts"       xargs -0 sha1sum -b >sha1sums"
               puts "(or use the --force switch)"
-              debug "temporary directory #{tmpdir} will be removed if it exists"
-              system "rm -rf #{tmpdir}" if File.directory?(tmpdir)
               exit -1
             end
           elsif not force_install
@@ -167,8 +152,6 @@ barclamps.values.sort_by{|v| v[:order]}.each do |bc|
             puts "  cd \"#{target}\"; find -type f -not -name sha1sums -print0 | \\"
             puts"       xargs -0 sha1sum -b >sha1sums"
             puts "(or use the --force switch)"
-            debug "temporary directory #{tmpdir} will be removed if it exists"
-            system "rm -rf #{tmpdir}" if File.directory?(tmpdir)
             exit -1
           end
         end
@@ -201,11 +184,6 @@ barclamps.values.sort_by{|v| v[:order]}.each do |bc|
       raise e
     end
   rescue StandardError => e
-    if ENV['DEBUG'] === 'true'
-      debug "temporary directory #{tmpdir} will be left for debugging if it exists"
-    else
-      rm_tmpdir(tmpdir)
-    end
     puts e
     puts e.backtrace
     puts "Install of #{bc[:name]} failed."
@@ -216,7 +194,5 @@ end
 generate_navigation
 generate_assets_manifest
 catalog
-
-rm_tmpdir(tmpdir)
 
 exit 0
--- a/scripts/barclamp_install.rb
+++ b/scripts/barclamp_install.rb
@@ -54,10 +54,9 @@ end
 
 usage if ARGV.length < 1
 
-tmpdir = "/tmp/bc_install-#{Process.pid}-#{Kernel.rand(65535)}"
+tmpdir = "/tmp/component_install-#{Process.pid}-#{Kernel.rand(65535)}"
 debug "tarball tmpdir: #{tmpdir}"
 Dir.mkdir(tmpdir)
-candidates = Array.new
 
 def rm_tmpdir(tmpdir)
   if File.directory?(tmpdir)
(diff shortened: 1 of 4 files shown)


scripts/install-chef-suse.sh
```diff
@@ -774,31 +774,20 @@ if [ -n "$CROWBAR_FROM_GIT" ]; then
...
-    cisco-ucs hyperv heat ceilometer manila trove tempest ; do
-    if test -d $BARCLAMP_SRC/$i; then
-        /opt/dell/bin/barclamp_install.rb $BARCLAMP_INSTALL_OPTS $BARCLAMP_SRC/$i
-    fi
-done
+/opt/dell/bin/barclamp_install.rb $BARCLAMP_INSTALL_OPTS $required_components
 
+# Install optional components if they're present
+if test -d $BARCLAMP_SRC/hyperv; then
+    /opt/dell/bin/barclamp_install.rb $BARCLAMP_INSTALL_OPTS $i
```
vuntz: You're not doing a loop here, so `$i` won't work ;-)

scripts/barclamp_mgmt_lib.rb
```diff
@@ -488,11 +455,6 @@ def bc_install_layout_1_app(from_rpm, bc, bc_path, yaml)
     FileUtils.rm manifest_created rescue nil
   end
 
-  filelist = File.join BARCLAMP_PATH, "#{bc}-filelist.txt"
-  File.open( filelist, 'w' ) do |out|
-    files.each { |line| out.puts line }
-  end
```
vuntz: Note that the uninstall command requires this file to exist (so it can remove files that were copied by the install script).

scripts/barclamp_mgmt_lib.rb
```diff
@@ -114,18 +93,6 @@ def catalog(bc_path)
...
-      [ File.join(bc_path, '..', name, 'chef', 'data_bags', 'crowbar', "bc-template-#{name}.json"), \
-        File.join(bc_path, '..', "barclamp-#{name}", 'chef', 'data_bags', 'crowbar', "bc-template-#{name}.json"), \
-        File.join(bc_path, '..', '..', 'chef', 'data_bags', 'crowbar', "bc-template-#{name}.json") ].each do |f|
-        next unless File.exist? f
-        s = JSON::load File.open(f, 'r')
-        description = s['description'] unless s.nil?
-        break if description
-      end
-    end
-    # template = File.join bc_path, name,
```
vuntz: I'm fine dropping that, but we first need to fix the yml files that don't have a description. I could find nfs_client, ntp and suse-manger-client just in crowbar-core. I would also add a big nice warning if it's nil.

scripts/barclamp_install.rb
```diff
@@ -54,10 +54,13 @@ def usage()
...
 
-tmpdir = "/tmp/bc_install-#{Process.pid}-#{Kernel.rand(65535)}"
+tmpdir = "/tmp/component_install-#{Process.pid}-#{Kernel.rand(65535)}"
 debug "tarball tmpdir: #{tmpdir}"
 Dir.mkdir(tmpdir)
-candidates = Array.new
+
+log_path = File.join '/var', 'log', 'crowbar'
+FileUtils.mkdir log_path unless File.directory? log_path
+log = File.join log_path, "component_install_#{Time.now.strftime("%Y%m%dT%H%M")}.log"
```
vuntz: Hrm, I really dislike that. Just append to a log file without a date in the filename, otherwise we'll end up with thousands of such files.

Also we need to make sure to have the correct logrotate rule.

scripts/barclamp_install.rb
```diff
@@ -54,10 +54,13 @@ def usage()
 
 usage if ARGV.length < 1
 
-tmpdir = "/tmp/bc_install-#{Process.pid}-#{Kernel.rand(65535)}"
+tmpdir = "/tmp/component_install-#{Process.pid}-#{Kernel.rand(65535)}"
```
vuntz: I know it's not your fault, but can we use some real solution instead of a hacky thing on this line (mktemp)?

scripts/barclamp_install.rb
```diff
@@ -188,7 +189,13 @@ def rm_tmpdir(tmpdir)
     end
     debug "installing barclamp"
     begin
-      bc_install from_rpm, bc[:name], bc[:src], bc[:yaml]
+      if bc[:yaml]["crowbar"]["layout"].to_i == 1
+        debug "Installing app components"
+        bc_install_layout_1_app from_rpm, bc[:name], bc[:src], bc[:yaml]
+        bc_install_layout_1_chef_migrate bc[:name], bc[:migrate], log
+      else
+        raise "ERROR: could not install barclamp #{bc[:name]} because #{bc[:yaml][:barclamp][:crowbar_layout]} is unknown layout."
```
vuntz: Should we really `raise`? Since some barclamps will have been installed, maybe we should only output an error and go on?

scripts/barclamp_mgmt_lib.rb
```diff
@@ -468,11 +435,11 @@ def bc_install_layout_1_app(from_rpm, bc, bc_path, yaml)
   end
 
   # copy over the crowbar YAML file, needed to update catalog
-  yml_barclamp = get_crowbar_yml_path(bc_path)
+  yml_barclamp = get_yml_paths(bc_path, bc)
   yml_path = File.join CROWBAR_PATH, 'barclamps'
   yml_created = File.join(yml_path, "#{bc}.yml")
   FileUtils.mkdir yml_path unless File.directory? yml_path
-  FileUtils.cp yml_barclamp, yml_created
+  FileUtils.cp yml_barclamp, yml_path
```
vuntz: Not sure I understand this line change.

vuntz: Oh, is it because it's copying multiple files? In which case `yml_barclamp` should probably be `yml_barclamps`.

scripts/barclamp_install.rb
```diff
@@ -66,47 +69,42 @@ def rm_tmpdir(tmpdir)
...
-    candidates << File.join(BARCLAMP_PATH, src)
+  if Dir.exist?(src)
+    barclamp_yml_files += get_yml_paths(src)
+    component_paths.push src
+  elsif Dir.exist?(File.join(BARCLAMP_PATH, src))
+    barclamp_yml_files += get_yml_paths(File.join(BARCLAMP_PATH, src))
+    component_paths.push File.join(BARCLAMP_PATH, src)
   else
-    puts "#{src} is not a barclamp, ignoring."
+    puts "#{srv} is not a valid component name, skiping."
```
vuntz: `srv` looks like a typo.

vuntz: s/skiping/skipping/ btw :-)

scripts/barclamp_install.rb
```diff
@@ -66,47 +69,42 @@ def rm_tmpdir(tmpdir)
...
-    # We were handed something that looks like a path to a barclamp
-    candidates << File.expand_path(src)
-  when !get_crowbar_yml_path(File.join(BARCLAMP_PATH, src)).nil?
-    candidates << File.join(BARCLAMP_PATH, src)
+  if Dir.exist?(src)
+    barclamp_yml_files += get_yml_paths(src)
+    component_paths.push src
+  elsif Dir.exist?(File.join(BARCLAMP_PATH, src))
+    barclamp_yml_files += get_yml_paths(File.join(BARCLAMP_PATH, src))
+    component_paths.push File.join(BARCLAMP_PATH, src)
```
vuntz: There's something I'm not getting: if I do `barclamp_install.rb core` it will look in /opt/dell/barclamps/core. But the yml files are still in /opt/dell/barclamps/{crowbar,dns,ntp,...}/. How can this work? I assume I'm missing something :-)

scripts/barclamp_mgmt_lib.rb
```diff
@@ -523,51 +485,47 @@ def bc_install_layout_1_app(from_rpm, bc, bc_path, yaml)
...
-  ensure
-    temp.close!
+def bc_install_layout_1_chef(from_rpm, component_paths, log)
+  components = Array.new
+  component_paths.each do |component_path|
+    components.push(File.basename(component_path))
   end
 
-  do_migrate = check_schema_migration(bc)
+  File.open(log, "a") { |f| f.puts("======== Installing chef componets -- #{Time.now.strftime('%c')} ========") }
```
vuntz: Typo (componets)

scripts/barclamp_install.rb
```diff
@@ -188,7 +189,13 @@ def rm_tmpdir(tmpdir)
     end
     debug "installing barclamp"
     begin
-      bc_install from_rpm, bc[:name], bc[:src], bc[:yaml]
+      if bc[:yaml]["crowbar"]["layout"].to_i == 1
+        debug "Installing app components"
+        bc_install_layout_1_app from_rpm, bc[:name], bc[:src], bc[:yaml]
+        bc_install_layout_1_chef_migrate bc[:name], bc[:migrate], log
```
vuntz: Can't we simply do `bc_install_layout_1_chef_migrate(bc[:name], log) if bc[:migrate]`?

scripts/barclamp_install.rb
rsalevsky: Yep, we added a workaround to macro in the crowbar package. When this is removed the yml files are in the crowbar,dns,ntp,... folders.

scripts/barclamp_install.rb
rsalevsky: Just copied the code but you are right the raise makes no sense.

scripts/barclamp_install.rb
vuntz: I assume you mean that when removed, they will be in the core folder. Ok, thanks!

scripts/barclamp_mgmt_lib.rb
rsalevsky: Warning is implemented and PR is open

rsalevsky:
The barclamp_uninstall script needs also some fixes now. Will look at this on Monday.

rsalevsky:
@vuntz Tested all the stuff today, works sand is really fast unlike before.

scripts/barclamp_install.rb
```diff
@@ -54,10 +55,12 @@ def usage()
 
 usage if ARGV.length < 1
 
-tmpdir = "/tmp/bc_install-#{Process.pid}-#{Kernel.rand(65535)}"
+tmpdir = Dir.mktmpdir "component-install-"
```
vuntz: Apologies for my earlier comment, but I suddenly realize that we don't use `tmpdir` at all? So maybe we should just drop all mentions of it?

scripts/barclamp_uninstall.rb
```diff
@@ -49,26 +49,39 @@ def usage()
...
+      barclamp = YAML.load_file yml_file
+
+      if barclamp["barclamp"] and barclamp["barclamp"]["name"]
+        name = barclamp["barclamp"]["name"]
+      else
+        puts "Barclamp at #{yml_file} has no name, skipping"
+        next
+      end
+
+      bc = barclamp["barclamp"]["name"].chomp.strip
```
vuntz: I guess you wanted to use the `name` variable here? Or maybe change the block above from `if` to `unless` (with no `else`)?

vuntz:
Didn't see anything obviously wrong, so I'd be +1 now. I guess we can't have proper gating for it?

scripts/barclamp_install.rb
tboerger: Yes, please remove all the `tmpdir` parts... They are totally useless :)
Especially the remove instructions for tmpdir as it's `Dir.mktmpdir` anyway now.

tboerger:
:+1:

rsalevsky:
Let's merge, tested it with patched packages and several mkcloud runs.

PR #2151, opened 2015-09-18
Skip SES2 update checks

dirkmueller:
Update channel doesn't exist right now.


aplanas:
+1 (Potentially fixing Jenkin's job 7925)

vuntz:
+1

vuntz:
Actually, I think the issue is only for the Updates repo, isn't it?

dirkmueller:
@vuntz right. removed pool and added a comment.

PR #2154, opened 2015-09-21
install-chef-suse: Switch to SLE12 SP1

vuntz:
This should be merged only when all the code has been ported to SP1, as this removes support for running cloud services on SLE12 GA.


vuntz:
https://github.com/SUSE-Cloud/automation/pull/559 and https://github.com/crowbar/crowbar-core/pull/42 can be used to play with this.

scripts/install-chef-suse.sh
```diff
@@ -511,20 +514,28 @@ cloud_dir=/srv/tftpboot/suse-12.0/repos/SLES12-Updates
 smt_dir=/srv/www/htdocs/repo/SUSE/Updates/SLE-SERVER/12/x86_64/update
 test ! -e $cloud_dir -a -d $smt_dir && ln -s $smt_dir $cloud_dir
 
-cloud_dir=/srv/tftpboot/suse-12.0/repos/SUSE-OpenStack-Cloud-6-Pool
```
toabctl: Why do you remove ```SUSE-OpenStack-Cloud-6-Pool```? Don't we need this symlink?

vuntz: It's not removed, see line 525.

toabctl:
Don't know why but while testing the patch I had the case that the target platform of node2 was set to ```SLE12``` (but should be ```SLES 12 SP1```).  Forget and restart the node, target platform was set to ```SLES 12 SP1```

toabctl:
Tested manually, worked for me. +1

toabctl:
@vuntz is there still a reason why this PR has the "do not merge yet" label?

vuntz:
@toabctl in short, it can be merged when we're ready to switch to SP1. So this is up to you and @dirkmueller :-)

toabctl:
@vuntz first step would be to remove the label :-)

dirkmueller:
Merging..

PR #2157, opened 2015-09-28
Added new codeclimate config

tboerger:


rsalevsky:
+1

aspiers:
Thanks a lot for doing this!

+0.5 - this config uses Code Climate Classic but I think we need the new Platform with the Ruby engine otherwise none of our PRs will get tested.  http://docs.codeclimate.com/article/213-github-pull-request-integration#requirements says

> we currently do NOT support:
> [...]
> - Pull requests from forked repositories. (available only on the new Platform)

tboerger:
> +0.5 - this config uses Code Climate Classic but I think we need the new Platform with the Ruby engine otherwise none of our PRs will get tested. http://docs.codeclimate.com/article/213-github-pull-request-integration#requirements says

I don't know where you get this. The `.codeclimate.yml` has been introduced with the new platform. And on `crowbar-core` anatytics page you can already see:

> Warning! You've enabled the Code Climate Platform but don't have a codeclimate.yml yet. You can start with the example below.

And of course it's not available yet as it's not merged to master.

MaximilianMeister:
:+1:

tboerger:
Got enough votes and config is exactly like the example from Codelimate, merging.

aspiers:
> I don't know where you get this.

From here: http://docs.codeclimate.com/article/289-configuring-your-repository-via-codeclimate-yml

> The .codeclimate.yml has been introduced with the new platform

That seems to contradict that page, which says "If you'd like to enable our beta Engines-based Platform, configure your .codeclimate.yml file differently than the sample file above." and also says it's necessary to include `engines` and `ratings` sections in `.codeclimate.yml`.  However it then goes on to say:

> Important: After opting into our new Platform, for now you can switch back to our original Ruby, PHP, JavaScript and PHP analysis. To do so, remove the engines: and ratings: nodes from your .codeclimate.yml file, and also add a languages: node (see the sample file at the top of this help article).

So I suspect that what you've done so far will not result in any PRs from forks getting tested.  But we'll see.

PR #2158, opened 2015-10-01
install-chef-suse: Stop some services before reinstalling (bsc#928183)

vuntz:
This is needed so that we're sure they'll be started with the right
config (and won't fail to restart in a notification due to some chef
crash).

The services in question are apache2, dhcpd, nfsserver, xinetd.

In theory, we should also stop dnsmasq, named and ntpd, but we don't
want to lose DNS and ntpd might be queried for its configuration later
on.

https://bugzilla.suse.com/show_bug.cgi?id=928183


kbaikov:
+1

nkrinner:
+1

PR #2159, opened 2015-10-05
Move sanity checks to separate crowbar lib

rsalevsky:
Depends on https://github.com/crowbar/crowbar-core/pull/61

### DIFF
--- a/scripts/install-chef-suse.sh
+++ b/scripts/install-chef-suse.sh
@@ -304,31 +304,6 @@ case "$rootpw" in
         ;;
 esac
 
-# It is exceedingly important that 'hostname -f' actually returns an FQDN!
-# if it doesn't, add an entry to /etc/hosts, e.g.:
-#    192.168.124.10 cb-admin.example.com cb-admin
-if ! FQDN=$(hostname -f 2>/dev/null); then
-    die "Unable to detect fully-qualified hostname. Aborting."
-fi
-
-if ! DOMAIN=$(hostname -d 2>/dev/null); then
-    die "Unable to detect DNS domain name. Aborting."
-fi
-
-if [ -z "$FQDN" -o -z "$DOMAIN" ]; then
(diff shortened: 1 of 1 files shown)


tboerger:
LGTM

vuntz:
FWIW, I don't think this can be merged until the install process is: "open your browser and click the next buttons".

rsalevsky:
Depends on: https://github.com/crowbar/crowbar-core/pull/61

MaximilianMeister:
assuming it works :+1:  :)

vuntz:
> FWIW, I don't think this can be merged until the install process is: "open your browser and click the next buttons".

Would it be possible to add a rake task to call the checks? And then, while we still depend on install-suse-cloud, we would simply call the rake task here.

vuntz:
s/stat/start/ in commit message

scripts/install-chef-suse.sh
```diff
@@ -304,31 +304,6 @@ case "$rootpw" in
...
-if [ -z "$FQDN" -o -z "$DOMAIN" ]; then
-    die "Unable to detect fully-qualified hostname. Aborting."
-fi
-
-if ! resolved=$(getent ahosts $FQDN 2>/dev/null); then
-    die "Unable to resolve hostname $FQDN via host(1). Please check your configuration of DNS, hostname, and /etc/hosts. Aborting."
-fi
-
-IPv4_addr=$( echo "$resolved" | awk '{ if ($1 !~ /:/) { print $1; exit } }' )
-IPv6_addr=$( echo "$resolved" | awk '{ if ($1  ~ /:/) { print $1; exit } }' )
```
vuntz: -1: all the variables that were defined are actually used later in the script, so we need to keep them defined...

rsalevsky: Fixed, the IPv6 is not used.

jdsn:
:+1:

PR #2160, opened 2015-10-07
Require 'ses' component for storage-only installs

tserong:
Prior to the big crowbar repo merge, the SUSE Enterprise Storage package
was named 'crowbar-barclamp-suse-enterprise-storage'.  Now it's got a
nice short component name ('crowbar-ses'), so need to update the install
script to match.

### DIFF
--- a/scripts/install-chef-suse.sh
+++ b/scripts/install-chef-suse.sh
@@ -81,7 +81,7 @@ mkdir -p "`dirname "$LOGFILE"`"
 run_succeeded=
 
 is_ses () {
-    [ -d /opt/dell/barclamps/suse-enterprise-storage ]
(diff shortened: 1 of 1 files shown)


tserong:
Related question: crowbar-ses presently lives at https://github.com/tserong/crowbar-ses. It would be best if this were moved to either the crowbar org (because it's a crowbar component) or the SUSE org (because it's for a specific SUSE product), but I don't mind either way. What do other crowbarians think?

vuntz:
+1

(I feel it should be in the SUSE org, but don't have a super strong opinion about that; would be happy to let it live in crowbar org too)

rsalevsky:
+1

PR #2161, opened 2015-10-07
allow using old json files

bmwiedemann:
this helps our Cloud QA who use pre-defined .json files
for cloud4/5/6 testing
and can also help customers who want to setup a new cloud
with an old network.json

### DIFF
--- a/scripts/install-chef-suse.sh
+++ b/scripts/install-chef-suse.sh
@@ -390,6 +390,8 @@ if [ -n "$IPv4_addr" ]; then
     else
         NETWORK_JSON=/opt/dell/chef/data_bags/crowbar/template-network.json
     fi
+    # allow using old json files # TODO: drop in 2017
+    sed -i -e 's/bc-template-network/template-network/' $NETWORK_JSON
 
     if ! /opt/dell/bin/network-json-validator --admin-ip "$IPv4_addr" $NETWORK_JSON; then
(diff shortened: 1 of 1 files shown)


scripts/install-chef-suse.sh
```diff
@@ -390,6 +390,8 @@ if [ -n "$IPv4_addr" ]; then
     else
         NETWORK_JSON=/opt/dell/chef/data_bags/crowbar/template-network.json
     fi
+    # allow using old json files # TODO: drop in 2017
+    sed -i -e 's/bc-template-network/template-network/' $NETWORK_JSON
```
dirkmueller: can we add a bit more context to the sed so that it doesn't accidentally match anything else?

vuntz: Well, there really sould be no bc-template-network anywhere anymore, so not sure it's strictly needed...

vuntz:
+1

scripts/install-chef-suse.sh
bmwiedemann: I'd prefer to not include quotes, because those might be single or double-quotes and would over-complicate the matter

rhafer:
+1

dguitarbite:
Can you please update the branch?

bmwiedemann:
rebased

PR #2162, opened 2015-10-07
install-chef-suse: Add --debug switch

toabctl:
When running in debug mode, barclamp_install.rb is
much more verbose and also chef-client runs are providing
debugging infos.

### DIFF
--- a/scripts/install-chef-suse.sh
+++ b/scripts/install-chef-suse.sh
@@ -30,7 +30,7 @@ set -e
 usage () {
     # do not document --from-git option; it's for developers only
     cat <<EOF
-`basename $0` [-h|--help] [-v|--verbose]
+`basename $0` [-h|--help] [-v|--verbose] [-d|--debug]
 
 Install Crowbar on administration server.
 EOF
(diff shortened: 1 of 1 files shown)


scripts/install-chef-suse.sh
```diff
@@ -71,6 +78,8 @@ if [ -n "$CROWBAR_FROM_GIT" ]; then
             install -p -m 0755 $tool /opt/dell/bin/
         fi
     done
+else
+    BARCLAMP_INSTALL_OPTS="$BARCLAMP_INSTALL_OPTS --rpm"
```
vuntz: Hrm, did we lose the --rpm argument at some point?

vuntz:
+1

scripts/install-chef-suse.sh
toabctl: Nope. ```BARCLAMP_INSTALL_OPTS``` was just overwritten if installed from git. should be fine.

vuntz: Sorry, I didn't mean "lose in the patch", but we had it in the past and it seems we don't have it right now (until before your patch)?

toabctl: We had it, just not for the "install from git" case

rsalevsky:
+1

tboerger:
:+1:

dirkmueller:
+1

PR #2163, opened 2015-10-12
Run rake in production environment

rsalevsky:
Since the proposals are in the database we need to run the rake task for
migrating the proposals in the production env.


vuntz:
+1

tboerger:
Looks fine! :+1:

dirkmueller:
+1

PR #2166, opened 2015-10-19
Codeclimate/HoundCI configuration

tboerger:
* Added the same HoundCI configurations as we already have for all other crowbar components
* Updated the Codeclimate configuration to the same as all other crowbar components are following


ovisan:
:+1:

MaximilianMeister:
:+1:

LICENSE
```diff
@@ -1,5 +1,4 @@
-
-                                 Apache License
+Apache License
```
vuntz: I don't understand the changes on this file. Where does this come from?

PR #2168, opened 2015-10-21
Enable deletion of rsynced files within Guard

tboerger:


tboerger:
This deletes too much files. Closing this for now.

PR #2169, opened 2015-10-21
Improved development environment setup and guard sync

tboerger:
As guard-remote-sync doesn't support custom options like `chown` I'm forced to switch to the `cli_options` config.


MaximilianMeister:
thanks for updating this. i assume you tested it, because i haven't yet

:+1:

PR #2172, opened 2015-10-26
activate provisioner repos

MaximilianMeister:
this would replace https://github.com/SUSE-Cloud/automation/pull/625


tboerger:
Why did we integrate the activate/deactivate functionality anyway when everything gets activated on install?

scripts/install-chef-suse.sh
```diff
@@ -1166,6 +1166,18 @@ for s in xinetd dhcpd apache2 ; do
...
 done
 
+# activate provisioner repos
+if ! is_ses; then
+    for i in SLES12-SP1-Pool SLES12-SP1-Updates SUSE-OpenStack-Cloud-6-Pool SUSE-OpenStack-Cloud-6-Updates SLES12-SP1-HA-Pool SLES12-SP1-HA-Updates; do
+      curl --digest -u crowbar:crowbar -X POST -d "platform=suse-12.1" -d "repo=$i" http://$IP/utils/repositories/activate
+    done
+fi
+
+if [ -e $MEDIA/boot/x86_64/common ]; then
```
vuntz: `MEDIA` can be set to different things at different point in time, might be best to be more explicit and use the SLE12 directory directly.

scripts/install-chef-suse.sh
```diff
@@ -592,13 +592,6 @@ if ! is_ses; then
             /srv/tftpboot/suse-12.1/repos/Cloud \
             #1558be86e7354d31e71e7c8c2574031a
     fi
-
-    check_repo_tag repo    12.1 SLES12-SP1-Pool                     'obsproduct://build.suse.de/SUSE:SLE-12-SP1:GA/SLES/12.1/POOL/x86_64'
-    check_repo_tag repo    12.1 SLES12-SP1-Updates                  'obsrepository://build.suse.de/SUSE:Updates:SLE-SERVER:12-SP1:x86_64/update'
-    check_repo_tag repo    12.1 SUSE-OpenStack-Cloud-6-Pool         'obsproduct://build.suse.de/SUSE:SLE-12:SP1:Products:Cloud6/suse-openstack-cloud/6/POOL/x86_64' $REQUIRE_CLOUD
-    check_repo_tag repo    12.1 SUSE-OpenStack-Cloud-6-Updates      'obsrepository://build.suse.de/SUSE:Updates:OpenStack-Cloud:6:x86_64/update' $REQUIRE_CLOUD
-    check_repo_tag repo    12.1 SLE12-SP1-HA-Pool                   'obsproduct://build.suse.de/SUSE:SLE-12-SP1:GA/sle-ha/12.1/POOL/x86_64' 'false'
-    check_repo_tag repo    12.1 SLE12-SP1-HA-Updates                'obsrepository://build.suse.de/SUSE:Updates:SLE-HA:12-SP1:x86_64/update' 'false'
```
vuntz: Can you also drop the lines defining `REQUIRE_STORAGE`/`REQUIRE_CLOUD`, since they're unused?

scripts/install-chef-suse.sh
```diff
@@ -1166,6 +1154,18 @@ for s in xinetd dhcpd apache2 ; do
     fi
 done
 
+# activate provisioner repos
+if ! is_ses; then
+    for i in SLES12-SP1-Pool SLES12-SP1-Updates SUSE-OpenStack-Cloud-6-Pool SUSE-OpenStack-Cloud-6-Updates SLES12-SP1-HA-Pool SLES12-SP1-HA-Updates; do
+      curl --digest -u crowbar:crowbar -X POST -d "platform=suse-12.1" -d "repo=$i" http://$IP/utils/repositories/activate
```
vuntz: The admin password might not be `crowbar` (nor the username, btw). You need to use the `machine-install` user and its password (look in the script, you'll find it).

scripts/install-chef-suse.sh
```diff
@@ -1166,6 +1154,18 @@ for s in xinetd dhcpd apache2 ; do
     fi
 done
 
+# activate provisioner repos
+if ! is_ses; then
+    for i in SLES12-SP1-Pool SLES12-SP1-Updates SUSE-OpenStack-Cloud-6-Pool SUSE-OpenStack-Cloud-6-Updates SLES12-SP1-HA-Pool SLES12-SP1-HA-Updates; do
```
vuntz: Would it be possible to just get the list of repos from the yaml file?

scripts/install-chef-suse.sh
MaximilianMeister: sure

scripts/install-chef-suse.sh
MaximilianMeister: hmm, would you use `json-read` for that? I would have to parse this string then:
```
attributes.crowbar.users={"crowbar"=>{"password"=>"crowbar"}}
```
right? or is there a more elegant way?

scripts/install-chef-suse.sh
MaximilianMeister: maybe we can use the ruby repositories library somehow for that?

scripts/install-chef-suse.sh
MaximilianMeister: actually we can skip authentification, if we request `localhost:3000` as it is running on the admin node

scripts/install-chef-suse.sh
tboerger: Maybe with the upcoming crowbar-client? ;)

scripts/repos-check-suse
```diff
@@ -38,27 +38,6 @@ skip_check_for_repo () {
     return 1
 }
 
-check_key_file () {
```
vuntz: This is still used by `check_media_content`.

scripts/install-chef-suse.sh
vuntz: In https://github.com/crowbar/crowbar-core/pull/102, I added a `activate_all` API that you can use instead of iterating over a list of repos.

scripts/repos-check-suse
MaximilianMeister: bad vim search... thanks

scripts/install-chef-suse.sh
MaximilianMeister: yes, that would be the one, but it doesn't work yet. but I can already make the change

scripts/install-chef-suse.sh
```diff
@@ -563,13 +563,6 @@ REPOS_SKIP_CHECKS+=" SUSE-OpenStack-Cloud-6-Pool SUSE-OpenStack-Cloud-6-Updates"
 # FIXME: Update repo doesn't exist yet, can't be checked
 REPOS_SKIP_CHECKS+=" SUSE-Enterprise-Storage-2-Updates"
```
vuntz: I just realize that you can also remove the 5 lines above. I'm pretty sure we also have code that sets REPOS_SKIP_CHECKS elsewhere that can be dropped.

vuntz:
> Why did we integrate the activate/deactivate functionality anyway when everything gets activated on install?

Because repos can be added after the install... That's the whole point of the change.

scripts/install-chef-suse.sh
tboerger: > actually we can skip authentification, if we request localhost:3000 as it is running on the admin node

Even with authentication you can use the key from /etc/crowbar.install.key

vuntz:
So you still need to remove line 342 and the block starting at line 362...

vuntz:
+1

tboerger:
:+1:

rsalevsky:
+1

PR #2173, opened 2015-10-27
install-chef-suse: Re-run knife if /root/.chef/root.pem is missing

vuntz:
On re-run of install-chef-suse.sh, if we have /root/.chef/knife.rb but
not /root/.chef/root.pem, then we assume that knife is working while it
actually needs the certificate. So also check for existence of
/root/.chef/root.pem.


rsalevsky:
+1

jsuchome:
+1

vuntz:
Gating is red, but in cct, so really unrelated to this change => merging.

PR #2174, opened 2015-10-28
Reverted wrong changes to license content

tboerger:


aplanas:
+1 Changes compatible with http://www.apache.org/licenses/LICENSE-2.0.txt

PR #2176, opened 2015-11-02
Ignore mkcloud garbage

tboerger:


rhafer:
I am not against this PR, but I wonder how do you get those files into the crowbar checkout in the first place?

tboerger:
Talked to @jdsn, will skip that.

aspiers:
See also https://github.com/SUSE-Cloud/automation/issues/224

PR #2177, opened 2015-11-03
Sync sledgehammer with what we ship for sleshammer

vuntz:


tboerger:
LGTM

vuntz:
As this is just a sync, and we do not use the bits from here, I'll just merge. We should consider using this as our source for the package, though...

PR #2178, opened 2015-11-04
Support running install-chef-suse from Crowbar

MaximilianMeister:
depends on https://github.com/crowbar/crowbar-core/pull/119

### DIFF
--- a/scripts/install-chef-suse.sh
+++ b/scripts/install-chef-suse.sh
@@ -244,6 +244,7 @@ exit_handler () {
     fi
 
     if [ -z "$run_succeeded" ]; then
+        post_fail_handler
         kill_spinner_with_failed
         cat <<EOF | pipe_show_and_log
 
@@ -259,7 +260,13 @@ EOF
 }
 
 trap exit_handler EXIT
+trap post_fail_handler INT
 
+post_fail_handler ()
+{
+    touch /var/lib/crowbar/install/crowbar-install-failed
+    rm -f /var/lib/crowbar/install/crowbar_installing
+}
 
 # Real work starts here
 # ---------------------
@@ -278,6 +285,19 @@ ensure_service_running () {
     fi
 }
 
+cleanup_steps () {
+    for i in pre_sanity_checks \
+         run_services \
+         initial_chef_client \
+         barclamp_install \
+         bootstrap_crowbar_setup \
+         apply_crowbar_config \
+         transition_crowbar \
+         chef_client_daemon \
+         post_sanity_checks; do
+         rm -f /var/lib/crowbar/${i}
+    done
+}
 
 if [ -f /opt/dell/crowbar_framework/.crowbar-installed-ok ]; then
     run_succeeded=already_before
@@ -291,6 +311,12 @@ then please remove the following file:
     /opt/dell/crowbar_framework/.crowbar-installed-ok
 EOF
     exit 1
+fi
+
+if [ -f /var/lib/crowbar/install/crowbar-install-failed ] || [ "$CROWBAR_WIZARD_MODE" ]; then
+    rm -f /var/lib/crowbar/install/crowbar-install-failed
+    cleanup_steps
+    sqlite3 /opt/dell/crowbar_framework/db/production.sqlite3 "delete from proposals; delete from proposal_queues; vacuum;"
 fi
 
 FQDN=$(hostname -f 2>/dev/null);
--- a/scripts/install-chef-suse.sh
+++ b/scripts/install-chef-suse.sh
@@ -29,6 +29,8 @@ set -e
 
 crowbar_install_dir=/var/lib/crowbar/install
 
+touch $crowbar_install_dir/crowbar_installing
+
 usage () {
     # do not document --from-git option; it's for developers only
     cat <<EOF
--- a/scripts/install-chef-suse.sh
+++ b/scripts/install-chef-suse.sh
@@ -623,7 +623,7 @@ if [ -n "$CROWBAR_FROM_GIT" ]; then
     # ubuntu admin node.
 fi
 
-touch /var/lib/crowbar/pre_sanity_checks
+touch /var/lib/crowbar/install/pre_sanity_checks
 
 
 # Starting services
@@ -678,7 +678,7 @@ for service in $services; do
     ensure_service_running chef-${service}
 done
 
-touch /var/lib/crowbar/run_services
+touch /var/lib/crowbar/install/run_services
 
 
 # Initial chef-client run
@@ -721,7 +721,7 @@ EOF
 
 $chef_client
 
-touch /var/lib/crowbar/initial_chef_client
+touch /var/lib/crowbar/install/initial_chef_client
 
 
 # Barclamp installation
@@ -776,7 +776,7 @@ if test -d $BARCLAMP_SRC/hyperv; then
     /opt/dell/bin/barclamp_install.rb $BARCLAMP_INSTALL_OPTS hyperv
 fi
 
-touch /var/lib/crowbar/barclamp_install
+touch /var/lib/crowbar/install/barclamp_install
 
 # First step of crowbar bootstrap
 # -------------------------------
@@ -806,7 +806,7 @@ $chef_client
 # OOC, what, if anything, is responsible for starting rainbows/crowbar under bluepill?
 ensure_service_running crowbar
 
-touch /var/lib/crowbar/bootstrap_crowbar_setup
+touch /var/lib/crowbar/install/bootstrap_crowbar_setup
 
 
 # Second step of crowbar bootstrap
@@ -1050,7 +1050,7 @@ done
 
 # BMC support?
 
-touch /var/lib/crowbar/apply_crowbar_config
+touch /var/lib/crowbar/install/apply_crowbar_config
 
 
 # Third step of crowbar bootstrap
@@ -1090,7 +1090,7 @@ done
 # OK, let looper_chef_client run normally now.
 rm /var/run/crowbar/deploying
 
-touch /var/lib/crowbar/transition_crowbar
+touch /var/lib/crowbar/install/transition_crowbar
 
 
 # Starting more services
@@ -1102,7 +1102,7 @@ echo_summary "Starting chef-client"
 chkconfig chef-client on
 ensure_service_running chef-client
 
-touch /var/lib/crowbar/chef_client_daemon
+touch /var/lib/crowbar/install/chef_client_daemon
 
 
 # Final sanity checks
@@ -1131,7 +1131,7 @@ for s in dhcpd apache2 ; do
     fi
 done
 
-touch /var/lib/crowbar/post_sanity_checks
+touch /var/lib/crowbar/install/post_sanity_checks
 
 # activate provisioner repos
 curl -X POST http://localhost:3000/utils/repositories/activate_all
--- a/scripts/install-chef-suse.sh
+++ b/scripts/install-chef-suse.sh
@@ -131,7 +131,7 @@ pipe_show_and_log () {
         rm -f $t
         dialog --clear >&3
     fi
-    tee -a /dev/fd/3 /dev/fd/4 > /dev/null
+    tee -a /dev/fd/4 >&3
 }
 
 # Draw a spinner so the user knows something is happening
--- a/scripts/install-chef-suse.sh
+++ b/scripts/install-chef-suse.sh
@@ -27,6 +27,8 @@
 
 set -e
 
+crowbar_install_dir=/var/lib/crowbar/install
+
 usage () {
     # do not document --from-git option; it's for developers only
     cat <<EOF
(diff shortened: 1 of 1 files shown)


scripts/install-chef-suse.sh
```diff
@@ -292,12 +300,27 @@ EOF
...
+    for i in pre_sanity_checks \
+         run_services \
+         initial_chef_client \
+         barclamp_install \
+         bootstrap_crowbar_setup \
+         apply_crowbar_config \
+         transition_crowbar \
+         chef_client_daemon \
+         post_sanity_checks; do
+         [[ -f /opt/dell/.${i} ]] && rm /opt/dell/.${i}
```
vuntz: Missing crowbar_framework in the path (I assume).

vuntz: Ah, actually, you're really putting things directly in /opt/dell. That feels bad, tbh. Can't we just use /var/lib/crowbar instead? And no need to use dot files.

MaximilianMeister: done

rsalevsky:
5b3c87e and 2d12f3c has partly the same content

MaximilianMeister:
> 5b3c87e and 2d12f3c has partly the same content

@rsalevsky I did a fixup

tboerger:
I dislike all these separate files. I would prefer a single file, it's easy to cleanup, it's easy to read and it's easy to write.

Writing it with `echo run_services >> /opt/dell/crowbar_framework/tmp/status`, cleanup with `rm -f /var/lib/crowbar/status`, and reading with `Rails.root.join("tmp", "status").readlines.to_json`.

MaximilianMeister:
> I dislike all these separate files. I would prefer a single file, it's easy to cleanup, it's easy to read and it's easy to write.

> Writing it with echo run_services >> /opt/dell/crowbar_framework/tmp/status, cleanup with rm -f /var/lib/crowbar/status, and reading with Rails.root.join("tmp", "status").readlines.to_json

I personally don't care. we would loose the timestamp/modification time for debugging to determine how long each step took, so to compensate that we would also have to add a timestamp to each line in the file at least.

btw the line would be more like 
```
Rails.root.join("tmp", "status").readlines.map {|m| m.chomp}.to_json
```
in case that there is no timestamp

tboerger:
`Rails.root.join("tmp", "status").readlines.map(&:chomp).to_json` or `Rails.root.join("tmp", "status").readlines.map(&:strip).to_json`

vuntz:
Just curious: do we plan to move the content of the script in the rails app at a later step?

rsalevsky:
@vuntz Yes, there are some cards for this already.

rsalevsky:
+1 Gating finished install-chef-suse successful

tboerger:
I still got the opinion that a single file with a list of states makes more sense.

MaximilianMeister:
> I still got the opinion that a single file with a list of states makes more sense.

@rsalevsky @vuntz what are your opinions on this? can we agree on single or multi files?

rsalevsky:
@MaximilianMeister I have no strong opinion on that but somehow I prefer multi files.

vuntz:
> @rsalevsky @vuntz what are your opinions on this? can we agree on single or multi files?

No strong opinion either, as long as the file(s) is/are in a directory that makes sense.

scripts/install-chef-suse.sh
```diff
@@ -130,7 +131,7 @@ pipe_show_and_log () {
         rm -f $t
         dialog --clear >&3
     fi
-    tee -a /dev/fd/3 /dev/fd/4 > /dev/null
+    tee -a /dev/fd/4 >&3
```
vuntz: Hrm. I'm not a big fan of touching the dark magic here. Why would /dev/fd/3 go away, and why would this new way of writing to it work?

scripts/install-chef-suse.sh
```diff
@@ -252,13 +254,20 @@ You should also check the Installation Manual, in
 particular the Troubleshooting section.  Note that this script can
 safely be re-run multiple times if required.
 EOF
+        touch /opt/dell/crowbar_framework/.crowbar-install-failed
```
vuntz: No need for this, already done in `post_fail_handler`.

scripts/install-chef-suse.sh
```diff
@@ -277,6 +286,19 @@ ensure_service_running () {
...
+    for i in pre_sanity_checks \
+         run_services \
+         initial_chef_client \
+         barclamp_install \
+         bootstrap_crowbar_setup \
+         apply_crowbar_config \
+         transition_crowbar \
+         chef_client_daemon \
+         post_sanity_checks; do
+         rm -f /var/lib/crowbar/${i}
```
vuntz: Actually, if we use multiple files, maybe we should put this in /var/lib/crowbar/install/

scripts/install-chef-suse.sh
```diff
@@ -252,13 +254,20 @@ You should also check the Installation Manual, in
...
         kill_spinner
     fi
 }
 
 trap exit_handler EXIT
+trap post_fail_handler INT
 
+post_fail_handler ()
+{
+    touch /opt/dell/crowbar_framework/.crowbar-install-failed
```
vuntz: Can we move this to /var/lib/crowbar too?

scripts/install-chef-suse.sh
```diff
@@ -290,6 +312,10 @@ then please remove the following file:
     /opt/dell/crowbar_framework/.crowbar-installed-ok
 EOF
     exit 1
+elif [ -f /opt/dell/crowbar_framework/.crowbar-install-failed ] || [ "$CROWBAR_FIRST_INSTALL" ]; then
+    rm -f /opt/dell/crowbar_framework/.crowbar-install-failed
+    cleanup_steps
+    sqlite3 /opt/dell/crowbar_framework/db/production.sqlite3 "delete from proposals; delete from proposal_queues"
```
vuntz: Also, can't we have a better way of cleaning the database? Right now, every time we add a table, we'll have to update this, which is painful...

scripts/install-chef-suse.sh
vuntz: Oh, you actually create them there already ;-) I guess I gave that feedback earlier, but the update is missing here.

scripts/install-chef-suse.sh
```diff
@@ -252,13 +254,20 @@ You should also check the Installation Manual, in
...
     fi
 }
 
 trap exit_handler EXIT
+trap post_fail_handler INT
 
+post_fail_handler ()
+{
+    touch /opt/dell/crowbar_framework/.crowbar-install-failed
+    rm -f /var/lib/crowbar/crowbar_installing
```
vuntz: Missing /install in path.

scripts/install-chef-suse.sh
vuntz: Probably best to use a variable for the directory to avoid such issues.

scripts/install-chef-suse.sh
```diff
@@ -1093,6 +1130,10 @@ curl -X POST http://localhost:3000/utils/repositories/activate_all
 echo_summary "Installation complete!"
 
 touch /opt/dell/crowbar_framework/.crowbar-installed-ok
```
vuntz: Can we also move that one to /var/lib/crowbar/install/?

scripts/install-chef-suse.sh
```diff
@@ -42,6 +42,7 @@ while test $# -gt 0; do
         -h|--help|--usage|-\?) usage ;;
         -v|--verbose) CROWBAR_VERBOSE=1 ;;
         -d|--debug) CROWBAR_DEBUG=1 ;;
+        -c|--crowbar) CROWBAR_FIRST_INSTALL=1 && CROWBAR_VERBOSE=1 ;;
```
vuntz: s/CROWBAR_FIRST_INSTALL/CROWBAR_WIZARD_MODE/ ? Or something like this. It's not necessarily a first install :-)

scripts/install-chef-suse.sh
MaximilianMeister: I don't know who wrote that dark magic, but he should be able to find out.
I cannot explain that, sorry. the idea came by @bmwiedemann

scripts/install-chef-suse.sh
MaximilianMeister: then we should move `.crowbar-installed-ok` too, that's why i kept it there. you agree with that?

vuntz: Sure

MaximilianMeister: i see you have asked that later too

scripts/install-chef-suse.sh
vuntz: IIRC, we stole that from devstack. I'm just very reluctant to change something like this without understanding why this happens because this might mean we're missing another bug.

Did you check that this didn't break the script when installing without calling from crowbar? Also, can you reproduce the issue when manually calling the script with the same options as crowbar use?

MaximilianMeister: > Did you check that this didn't break the script when installing without calling from crowbar? Also, can you reproduce the issue when manually calling the script with the same options as crowbar use?

no not yet, i was working on the backup/restore mostly but i can hopefully do that tomorrow.

tboerger:
So the only benefit from your side about adding timestamps to measure the time or whatever is not added and not used. So there is not any benefit of multifile. Just the "downside" to manage even more files than a single one.

rsalevsky:
Okay, I agree with @tboerger we should use just a single file.

scripts/install-chef-suse.sh
```diff
@@ -612,6 +630,8 @@ if [ -n "$CROWBAR_FROM_GIT" ]; then
     # ubuntu admin node.
 fi
 
+echo "pre_sanity_checks, $(date)" >> $installation_steps
```
tboerger: I would prefer a space as a separator, easier to extract. Beside that the code smells like duplication, write a small func that just accepts the step :)

MaximilianMeister: `date` has spaces as well, that's why i used another separator

vuntz: `date -Iseconds`?

scripts/install-chef-suse.sh
```diff
@@ -287,11 +304,17 @@ Aborting: Administration Server is already deployed.
...
-    /opt/dell/crowbar_framework/.crowbar-installed-ok
+    $crowbar_install_dir/crowbar-installed-ok
 EOF
     exit 1
 fi
 
+if [ -f $crowbar_install_dir/crowbar-install-failed ] || [ "$CROWBAR_WIZARD_MODE" ]; then
+    rm -f $crowbar_install_dir/crowbar-install-failed
+    rm -f $installation_steps
+    sqlite3 /opt/dell/crowbar_framework/db/production.sqlite3 "delete from proposals; delete from proposal_queues"
```
tboerger: Please call vacuum at the end as well.

tboerger:
After adding vacuum to sqlite call you get my :+1:

rsalevsky:
:+1:

scripts/install-chef-suse.sh
jdsn: I deployed a M7 with only this patch and it successfully installed crowbar.
I can not explain the cause of the bug either, but I verified that this fixes this bug.

jdsn: So this means this patch is tested with the new and the old installer code path of https://github.com/SUSE-Cloud/automation/pull/654

scripts/install-chef-suse.sh
```diff
@@ -287,11 +304,17 @@ Aborting: Administration Server is already deployed.
 If you want to run the installation script again,
 then please remove the following file:
 
-    /opt/dell/crowbar_framework/.crowbar-installed-ok
+    $crowbar_install_dir/crowbar-installed-ok
 EOF
     exit 1
 fi
 
+if [ -f $crowbar_install_dir/crowbar-install-failed ] || [ "$CROWBAR_WIZARD_MODE" ]; then
```
vuntz: I asked earlier, but: Why not always do this? I don't see a case where we wouldn't want to do these things.

vuntz: Ah: is it because the sqlite database might not exist? If yes, then I would change the test accordingly.

scripts/install-chef-suse.sh
```diff
@@ -287,11 +304,17 @@ Aborting: Administration Server is already deployed.
...
+    rm -f $crowbar_install_dir/crowbar-install-failed
+    rm -f $installation_steps
+    sqlite3 /opt/dell/crowbar_framework/db/production.sqlite3 "delete from proposals; delete from proposal_queues; vacuum;"
```
vuntz: Can't we have a better way of cleaning the database? I suggest that we get a list of tables and delete everything for all of them except for the sessions table.

tboerger: Another option should be a reset of the database, e.g. `bin/rake db:migrate:reset`

tboerger: If we need to keep the `sessions` table we can migrate down to the `sessions` migration and migrate up again. That way we can be sure that we got a clean database.

PR #2179, opened 2015-11-04
Don't create secrets.env file while install-chef-suse run

rsalevsky:
Creating the secrets.env file is now part of the package.


tboerger:
:+1:

MaximilianMeister:
:+1:

vuntz:
@rsalevsky can we not do this in the package, but in the systemd service file of crowbar with `ExecStartPre`? That would make it easier for people to develop from git...

rsalevsky:
@vuntz I develop from git and there is no issue and you shouldn't use the service while developing anyway.

tboerger:
> @rsalevsky can we not do this in the package, but in the systemd service file of crowbar with ExecStartPre? That would make it easier for people to develop from git...

While you are working from git you should run the rails app in development mode, there you don't need this env variable anyway as the secrets.yml from crowbar already provides it. We need this env variable only for production environment.

PR #2180, opened 2015-11-06
Set the variables that are used later in the script

jsuchome:
They were accidentally removed by
https://github.com/crowbar/crowbar/pull/2159

Without them, IPv4_addr is not set, thus breaking chef-client configuration. DOMAIN and FQDN are  used in the script as well


MaximilianMeister:
aren't those checks in crowbar now? https://github.com/crowbar/crowbar-core/pull/61

rhafer:
Yes, but all the variables are used later or in install-suse-cloud

rhafer:
@jsuchome I would prefer if we just reverted the whole of #2159 and start over from there.

tboerger:
We should just define the variables. The checks are already done while starting crowbar.

PR #2182, opened 2015-11-06
Append always .title to navi i18n keys

tboerger:
With that change we avoid the conflict with our branding branch for the future.


rhafer:
+1 (We just have to make sure to fix the branding patch once more when merging this)

rsalevsky:
+1

tboerger:
The branding branch is totally fine (https://github.com/SUSE-Cloud/crowbar-core/blob/master-branding/crowbar_framework/config/locales/crowbar/en.yml#L82-L86). Who wants to update the patch? I would suggest @rsalevsky :)

Than we can merge this.

tboerger:
Merging, packages will be rebuilt now by @rsalevsky

PR #2181, opened 2015-11-06
Revert ae3555f0dfe8eda554543f80c0d0c383836a9a64

rhafer:
Reverts crowbar/crowbar#2159, which was merged prematurely. It causes 100% installation failures and turned all gating red.

PR #2183, opened 2015-11-06
Add missing install-chef-suse variables

rsalevsky:
These variables got accidentally removed by ae3555f.

Superseeds #2181 and #2180.

### DIFF
--- a/scripts/install-chef-suse.sh
+++ b/scripts/install-chef-suse.sh
@@ -292,6 +292,10 @@ EOF
     exit 1
 fi
 
+FQDN=$(hostname -f 2>/dev/null);
+DOMAIN=$(hostname -d 2>/dev/null);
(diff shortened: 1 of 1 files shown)


scripts/install-chef-suse.sh
```diff
@@ -292,6 +292,10 @@ EOF
     exit 1
 fi
 
+IPv4_addr=$( echo "$resolved" | awk '{ if ($1 !~ /:/) { print $1; exit } }' )
```
rhafer: ```$resolved``` still does not exist anywhere.

tboerger:
LGTM, gating fails because of tgt service (so it seems to be unrelated).

rhafer:
+1

jsuchome:
+1, but we have to remove current (reverting) patch from the package

rhafer:
@jsuchome I'll do that as soon as this one is merged.

PR #2184, opened 2015-11-07
Revert "Append always .title to navi i18n keys"

rsalevsky:
This reverts commit 194bc8d804ebf6af181d65b654d1de1527fbee62.

This part is broken afterwards:
```yml
  nav:
    nodes:
      title: 'Nodes'
      dashboard: 'Dashboard'
      batch: 'Bulk Edit'
      clusters: 'HA Clusters'
      roles: 'Active Roles'
      families: 'Family Groups'
    barclamps:
      title: 'Barclamps'
      all: 'All Barclamps'
      crowbar: 'Crowbar'
      queue: 'Deployment Queue'
    utils:
      title: 'Utilities'
      logs: 'Exported Items'
      repositories: 'Repositories'
    help:
      title: 'Help'
      docs: 'Documentation'
      crowbar_users: 'Crowbar Users Guide'
      wiki: 'Online Help'
```

fix would be to add the title everywhere like:

```yml
  nav:
    nodes:
      title: 'Nodes'
      dashboard:
        title: 'Dashboard'
```

but this seams not a good solution form me.

Revert is already in the package and branding patch is adapted.

### DIFF
--- a/scripts/barclamp_mgmt_lib.rb
+++ b/scripts/barclamp_mgmt_lib.rb
@@ -179,7 +179,7 @@ def prepare_navigation(hash, breadcrumb, indent, level)
       options_string.prepend(", ") unless options_string.empty?
 
       if values.keys.empty?
-        result.push "level#{level}.item :#{key}, t(\"nav.#{current_path.join(".")}.title\"), #{link}#{options_string}".indent(indent)
+        result.push "level#{level}.item :#{key}, t(\"nav.#{current_path.join(".")}\"), #{link}#{options_string}".indent(indent)
       else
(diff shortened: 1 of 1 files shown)


tboerger:
What about fixing it correctly instead of just reverting blind?

rsalevsky:
As I said for me this is not a good solution to add the title everywhere.

vuntz:
I'm fine reverting; afaik, the change really only matters because we want to help not break the branding patch (or was there any other reason?). And https://github.com/crowbar/crowbar-core/pull/121 would actually make this easy.

tboerger:
> As I said for me this is not a good solution to add the title everywhere.

Just check the level, if it's the first level it is safe to always append `.title`.

tboerger:
So let's revert it and get the docs update done really soon.

PR #2185, opened 2015-11-17
Drop support for SLES 12 repos

mjura:


tboerger:
:+1:

toabctl:
+1 (but that needs more changes in crowbar-core, right?)

vuntz:
+1

vuntz:
Actually: can you drop the bits for the SLE12 repos? We don't need them anymore.

mjura:
@vuntz : is it correct?

vuntz:
Nope, because now you don't create symlinks for SES 2.1 ;-)

mjura:
@vuntz: As you wish :) I fixed also path to directories

vuntz:
Good catch on the paths! Ideally, this would be a separate commit, though.

+1 after splitting the commit

vuntz:
Actually, I just realized there's still some code later on that is happening for SLE12 only (the checks for the install media). This should be fixed too.

vuntz:
check_or_create_ptf_repository and sign_repositories calls for 12.0 should be dropped too.

mjura:
I discovered also that we have separate check for SLE12 install media, because of ses

mjura:
All changes are in different commits

scripts/install-chef-suse.sh
```diff
@@ -452,95 +452,72 @@ supported_arches_ses="x86_64"
...
-    for arch in $supported_arches; do
-        # Checks for SLE12 SP1 medias
-        MEDIA=/srv/tftpboot/suse-12.1/$arch/install
-
-        # Only x86_64 is truly mandatory; other architectures are only checked
-        # if they exist
-        if [ ! -f $MEDIA/content -a $arch != "x86_64" ]; then
-            continue
-        fi
+for arch in $supported_arches; do
```
vuntz: You need something like:
```bash
if is_ses; then
  product_arches="$supported_arches_ses"
else
  product_arches="$supported_arches"
fi

for arch in $supported_arches; do
```

mjura: Ahhh, I haven't seen this. You are right. Thank you

scripts/install-chef-suse.sh
```diff
@@ -452,95 +452,78 @@ supported_arches_ses="x86_64"
...
     test ! -e $cloud_dir -a -d $smt_dir && ln -s $smt_dir $cloud_dir
   fi
 done
 
-if ! is_ses; then
-    for arch in $supported_arches; do
-        # Checks for SLE12 SP1 medias
-        MEDIA=/srv/tftpboot/suse-12.1/$arch/install
+if is_ses; then
+  product_arches="$supported_arches_ses"
```
dirkmueller: wrong indentation, should be 4 spaces

scripts/install-chef-suse.sh
```diff
@@ -452,95 +452,78 @@ supported_arches_ses="x86_64"
...
+else
+  product_arches="$supported_arches"
```
dirkmueller: wrong indentation, should be 4 spaces

scripts/install-chef-suse.sh
mjura: https://github.com/crowbar/crowbar/pull/2188

scripts/install-chef-suse.sh
mjura: https://github.com/crowbar/crowbar/pull/2188

PR #2187, opened 2015-11-20
Reenable bashate travis testing for PRs

dirkmueller:
Apparently this code refactoring was never tested.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -235,24 +235,23 @@ namespace :test do
 
   desc "Run bashate tests"
   task bashate: [:dependencies] do
-    script = "scripts/install-chef-suse.sh"
-    puts "checking #{script}"
+    ["scripts/install-chef-suse.sh",
+     "sledgehammer/start-up.sh"].each do |script|
+      system("bash -n #{script}")
+      if $?.exitstatus != 0
+        exit 3
+      end
 
-    system("bash -n #{script}")
-    if $?.exitstatus != 0
-      exit 3
+      system("bashate --ignore E010,E011,E020 #{script}")
+      if $?.exitstatus != 0
+        exit 4
+      end
+
+      # checking for tabs in the file
+      if File.open(script).grep(/\t/).any?
+        exit 5
+      end
     end
-
-    system("bashate --ignore E010,E011,E020 #{script}")
-    if $?.exitstatus != 0
-      exit 4
-    end
-
-    # checking for tabs in the file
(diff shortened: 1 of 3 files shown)


tboerger:
:+1:

rsalevsky:
+1

vuntz:
+1

aspiers:
> Ignore E006: Line too long …
>
> There are dozens of offenders in the code meanwhile, and got
> no time to fix them all. Sorry Adam :)

So when will this be fixed, if not now?  There are 72 long lines in these 2 scripts.  I really doubt it would take more than 10-20 minutes to fix, which IMHO is a far better approach than adding an ignore which would probably stay there for another 3 years.  But if you really think it will take hours to insert 72 newline characters into two files then please add a story to the backlog for it and assign it to grapes.  Otherwise -1 from me, because this ignore will allow the technical debt to get worse before it gets better.

dirkmueller:
@aspiers not all can be easily wrapped. I actually tried it and ended up running into areas where I need to test whether the constructs are behaving equivalent, and I just didn't have time for that. BTW, in case you missed it, E006 is also disabled in the bashate tests on other repos, so its not a one-off-thing.

dirkmueller:
@aspiers in addition, the bashate line length check is totally broken :-(

see https://review.openstack.org/#/c/248253/

aspiers:
> I actually tried it and ended up running into areas where I need to test whether the constructs are behaving equivalent, and I just didn't have time for that.

Ah OK.  So why not submit the fixes for the easy cases, if you already did them?  I'd be much more comfortable with this PR if there were only a few cases left to fix - even if they are a bit tricky, it makes it more likely that the remaining ones will be fixed soon.

It's unfortunate that Hound doesn't support shell, otherwise we could adopt an incremental approach to this.  However https://github.com/filib/codeclimate-shellcheck/ is another option which looks even more promising to me.

> BTW, in case you missed it, E006 is also disabled in the bashate tests on other repos, so its not a one-off-thing.

I don't see how that's relevant?  Bad code in one repo does not justify bad code in another :)

dirkmueller:
@aspiers done

aspiers:
There are still 34 E006 violations of which several look easily fixable to me.  However this is an improvement on before, so +0.2 from me.

toabctl:
+1

dirkmueller:
(created https://trello.com/c/IRHo5Bnx as a followup)

PR #2189, opened 2015-11-24
always cleanup the rails db

MaximilianMeister:
and use rake for cleaning up

### DIFF
--- a/scripts/install-chef-suse.sh
+++ b/scripts/install-chef-suse.sh
@@ -312,7 +312,10 @@ fi
 if [ -f $crowbar_install_dir/crowbar-install-failed ] || [ "$CROWBAR_WIZARD_MODE" ]; then
     rm -f $crowbar_install_dir/crowbar-install-failed
     rm -f $installation_steps
-    sqlite3 /opt/dell/crowbar_framework/db/production.sqlite3 "delete from proposals; delete from proposal_queues; vacuum;"
+    pushd .
+    cd /opt/dell/crowbar_framework
+    rake db:migrate:reset
+    popd
 fi
 
 FQDN=$(hostname -f 2>/dev/null);
--- a/scripts/install-chef-suse.sh
+++ b/scripts/install-chef-suse.sh
@@ -278,6 +278,15 @@ post_fail_handler ()
     rm -f $crowbar_install_dir/crowbar_installing
 }
 
+function reset_crowbar()
+{
+    rm -f $crowbar_install_dir/crowbar-install-failed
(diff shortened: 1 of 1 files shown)


scripts/install-chef-suse.sh
```diff
@@ -278,6 +278,16 @@ post_fail_handler ()
     rm -f $crowbar_install_dir/crowbar_installing
 }
 
+function rails_db_cleanup()
+{
+    rm -f $crowbar_install_dir/crowbar-install-failed
+    rm -f $installation_steps
+    pushd .
```
tboerger:
```
pushd /opt/dell/crowbar_framework > /dev/null
rake db:migrate:reset
popd  > /dev/null
```

tboerger:
A tiny improvement, afterwards you get my :+1:

scripts/install-chef-suse.sh
```diff
@@ -278,6 +278,15 @@ post_fail_handler ()
     rm -f $crowbar_install_dir/crowbar_installing
 }
 
+function rails_db_cleanup()
+{
+    rm -f $crowbar_install_dir/crowbar-install-failed
+    rm -f $installation_steps
```
vuntz: So these two bits are not really related to db cleanup, so maybe rename the function (reset_crowbar, maybe -- not super good name either, though)?

scripts/install-chef-suse.sh
```diff
@@ -278,6 +278,15 @@ post_fail_handler ()
...
+    pushd /opt/dell/crowbar_framework > /dev/null
+    rake db:migrate:reset
```
vuntz: Nice, thanks for doing that! But doesn't that remove the existing session from the db; is it an issue if installing from the webui?

mjura: Yeap, I worry about dropping whole database and recreating it once again

tboerger: As mentioned in the previous pull request we can migrate down to `sessions` table to be keep the session data working. An example will follow.

tboerger: https://github.com/crowbar/crowbar-core/pull/141

tboerger: You can call `bin/rake db:cleanup` with the mentioned pull request.

MaximilianMeister:
depends on https://github.com/crowbar/crowbar-core/pull/141

rsalevsky:
:+1:

tboerger:
Same functionality as before but with cleaner handling, merging.

PR #2190, opened 2015-11-24
Sync codeclimate config, remove bundler-audit

tboerger:
Remove bundler-audit check, it won't work as we are not providing any Gemfile.lock within git.


rsalevsky:
+1

PR #2192, opened 2015-11-25
Detect that puma is running

aplanas:
Detect that puma is running making queries to the local puma
instance.  This solution avoid errors 503 from apache, that
try to connect puma before it is running.

### DIFF
--- a/scripts/install-chef-suse.sh
+++ b/scripts/install-chef-suse.sh
@@ -295,6 +295,21 @@ ensure_service_running () {
     fi
 }
 
+wait_for_crowbar ()
+{
+    local crowbar_up=0
+    for ((x=1; x<10; x++)); do
+        nc -z localhost 3000 && {
+            crowbar_up=1
+            break
+        }
+        sleep 2
+    done
+    if [ $crowbar_up -ne 1 ]; then
+        die "Crowbar service is not running. Please check $LOGFILE."
+    fi
+}
+
 if [ -f $crowbar_install_dir/crowbar-installed-ok ]; then
     run_succeeded=already_before
 
@@ -948,6 +963,8 @@ touch /var/run/crowbar/deploying
 #   /opt/dell/bin/looper_chef_client.sh
 #
 # which exits immediately if /var/run/crowbar/deploying exists.
+
+wait_for_crowbar
 
 # From here, you should probably read along with the equivalent steps in
 # install-chef.sh for comparison


tboerger:
Why do we need that? Crowbar have to run with the new installer beforehand anyway?

scripts/install-chef-suse.sh
```diff
@@ -949,6 +967,8 @@ touch /var/run/crowbar/deploying
 #
 # which exits immediately if /var/run/crowbar/deploying exists.
 
+ensure_crowbar_running
```
vuntz: Hrm, but aren't we creating the "default" proposal for the crowbar barclamp just in the block below?

aplanas:
This PR can be the wrong solution for a real problem, that behaves like this in qa3:

* crowbar show proposal returns 503, instead the 'no default proposal' error or the JSON
* if we sleep for ~10 seconds, in the line where `ensure_crowbar_running`, the show proposal works properly
* if I iterate asking crowbar show proposal until I have something different to 503, the next create proposal or show proposal will cause again a 503

I do not see anything relevant in the usual logs (install.log or apache logs)

vuntz:
The 503 is just puma starting; I'm fine with your iteration in that case, but I would really just fetch / and not any proposal that doesn't exist at this point in time.

Another way to fix it is to fix the systemd service to not return until puma is done with the initialization. Not sure if it's desirable (or doable), though.

aplanas:
* Make the query directly to Puma
* Fix the description and the title of the PR

scripts/install-chef-suse.sh
```diff
@@ -295,6 +295,22 @@ ensure_service_running () {
...
+    local crowbar_up=0
+    local error="Errno::ECONNREFUSED"
+    for ((x=1; x<10; x++)); do
+        ($CROWBAR crowbar list --port 3000) 2>&1 >/dev/null | grep -q "$error" || {
+            crowbar_up=1
+            break
+        }
+        sleep 2
+    done
+    if [ -z crowbar_up ]; then
```
vuntz: `-z` doesn't work if you set the initial value to `0`, I think. Also missing `$`.

scripts/install-chef-suse.sh
```diff
@@ -295,6 +295,22 @@ ensure_service_running () {
     fi
 }
 
+ensure_crowbar_running ()
```
vuntz: I'd rename to `wait_for_crowbar`

aplanas:
* Rename to `wait_for_crowbar`
* Compare `$crowbar_up` explicitly to 1

vuntz:
+1

scripts/install-chef-suse.sh
```diff
@@ -295,6 +295,22 @@ ensure_service_running () {
     fi
 }
 
+wait_for_crowbar ()
+{
+    local crowbar_up=0
+    local error="Errno::ECONNREFUSED"
+    for ((x=1; x<10; x++)); do
+        ($CROWBAR crowbar list --port 3000) 2>&1 >/dev/null | grep -q "$error" || {
```
dirkmueller: set LC_ALL=C, otherwise grepping for the error message might not work

aplanas:
I was thinking to use `nc -z localhost 3000` as a test, instead of `crowbar crowbar proposal list`.  The first one is more generic, but maybe the second one is more explicit and clear

aplanas:
* Replace the call to `crowbar` and `grep` to `nc`
* Fix comparison when give up

aplanas:
I tested it with `set -e` and looks like that works

scripts/install-chef-suse.sh
```diff
@@ -295,6 +295,21 @@ ensure_service_running () {
     fi
 }
 
+wait_for_crowbar ()
+{
+    local crowbar_up=0
+    for ((x=1; x<10; x++)); do
+        nc -z localhost 3000 && {
```
vuntz: Style-wise, I would do `if nc -z localhost 3000; then`

But no big deal, so +1.

tboerger:
If we really need this :+1:

PR #2194, opened 2015-11-30
json-read: Fix output to be valid JSON for hash values

vuntz:
We were outputting ruby representation for attributes, which is not
valid JSON when the value is not a string / integer / boolean / etc.

### DIFF
--- a/scripts/json-read
+++ b/scripts/json-read
@@ -88,7 +88,11 @@ def main()
       fields = attr.gsub('\.', MAGIC_DOT).split('.')
       fields.each { |x| data = data[x.gsub(MAGIC_DOT, '.')] unless data.nil? }
 
-      puts "#{attr}=#{data}"
+      if data.is_a? Hash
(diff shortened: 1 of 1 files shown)


aspiers:
+1

MaximilianMeister:
:+1:

tboerger:
Why are we anyway writing json on our own? Ruby got methods to dump valid JSON

vuntz:
> Why are we anyway writing json on our own? Ruby got methods to dump valid JSON

I'm confused :-) Do you want to ask "why do we have this tool", or "why this change", or something else?

tboerger:
> I'm confused :-) Do you want to ask "why do we have this tool", or "why this change", or something else?

I'm asking why we are doing it so fragile and complicated :D

aspiers:
@tboerger PRs very welcome (and no, I don't mean [this](http://www.urbandictionary.com/define.php?term=patches+are+welcome) ;-) )

PR #2195, opened 2015-12-02
Various fixes / cleanups / simplifications for barclamp_install

vuntz:
Note that gating cannot be green as this requires some changes in the packaging (the yml files and manifests need to be put directly in the right location by the packages, which makes sense anyway)

### DIFF
--- a/scripts/barclamp_install.rb
+++ b/scripts/barclamp_install.rb
@@ -122,7 +122,7 @@ barclamp_yml_files.sort.each do |yml_file|
     next
   end
 
-  barclamps[name] = { :name => name, :migrate => do_chef && check_schema_migration(name) }
+  barclamps[name] = { name: name, migrate: do_chef && check_schema_migration(name) }
   debug "barclamp[#{name}] (from #{yml_file}) = #{barclamps[name].pretty_inspect}"
 end
 
--- a/scripts/barclamp_install.rb
+++ b/scripts/barclamp_install.rb
@@ -69,7 +69,7 @@ ARGV.each do |src|
     barclamp_yml_files += get_yml_paths(File.join(BARCLAMP_PATH, src))
     component_paths.push File.join(BARCLAMP_PATH, src)
   else
-    puts "#{src} is not a valid component name, skiping."
+    puts "#{src} is not a valid component name, skipping."
   end
 end
 
@@ -86,7 +86,7 @@ barclamp_yml_files.each do |yml_file|
     debug "trying to parse crowbar YAML file in #{yml_file}"
     barclamp = YAML.load_file yml_file
   rescue
-    puts "Exception occured while parsing crowbar YAML file in #{yml_file}, skiping"
+    puts "Exception occured while parsing crowbar YAML file in #{yml_file}, skipping"
     next
   end
 
--- a/scripts/barclamp_install.rb
+++ b/scripts/barclamp_install.rb
@@ -118,6 +118,9 @@ barclamps.values.sort_by{|v| v[:order]}.each do |bc|
   bc[:migrate] = check_schema_migration(bc[:name])
 end
 
+component_paths.each do |component_path|
+  bc_install_layout_1_app(from_rpm, component_path)
+end
 bc_install_layout_1_chef(from_rpm, component_paths, log)
 
 debug "installing barclamps:"
@@ -178,7 +181,6 @@ barclamps.values.sort_by{|v| v[:order]}.each do |bc|
     begin
       if bc[:yaml]["crowbar"]["layout"].to_i == 1
         debug "Installing app components"
-        bc_install_layout_1_app from_rpm, bc[:name], bc[:src]
         bc_install_layout_1_chef_migrate bc[:name], log if bc[:migrate]
       else
         debug "Could not install barclamp #{bc[:name]} because #{bc[:yaml][:barclamp][:crowbar_layout]} is unknown layout."
--- a/scripts/barclamp_install.rb
+++ b/scripts/barclamp_install.rb
@@ -84,7 +84,11 @@ ARGV.each do |src|
 
   next if component_dir.nil?
 
-  barclamp_yml_files += get_yml_paths(component_dir)
+  if from_rpm
+    barclamp_yml_files += get_yml_paths_from_rpm(File.basename(component_dir))
+  else
+    barclamp_yml_files += get_yml_paths(component_dir)
+  end
   component_paths.push component_dir
 end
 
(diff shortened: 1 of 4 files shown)


vuntz:
Note that I could split this in multiple PRs if preferred.

scripts/barclamp_install.rb
```diff
@@ -58,19 +55,41 @@ def usage()
 FileUtils.mkdir log_path unless File.directory? log_path
 log = File.join log_path, "component_install.log"
 
+install_lib_dir = "/var/lib/crowbar/install"
+installation_steps = File.join(install_lib_dir, "installation_steps")
+if File.exist? installation_steps
```
MaximilianMeister: just use `installation_steps.exist?`

scripts/barclamp_install.rb
```diff
@@ -58,19 +55,41 @@ def usage()
...
+  steps = File.readlines(installation_steps).map do |step|
```
MaximilianMeister: `installation_steps.readlines` should work too i think

scripts/barclamp_mgmt_lib.rb
```diff
@@ -210,8 +217,13 @@ def generate_navigation
...
   end
 
+  config_dir = File.join(CROWBAR_PATH, 'config')
+  unless File.directory? config_dir
+    FileUtils.mkdir_p config_dir
+  end
+
   File.open(
-    File.join(CROWBAR_PATH, 'config', 'navigation.rb'),
+    File.join(config_dir, 'navigation.rb'),
```
MaximilianMeister: `config_dir.join('navigation.rb')`

scripts/barclamp_mgmt_lib.rb
```diff
@@ -210,8 +217,13 @@ def generate_navigation
     current.deep_merge! config["nav"]
   end
 
+  config_dir = File.join(CROWBAR_PATH, 'config')
+  unless File.directory? config_dir
+    FileUtils.mkdir_p config_dir
+  end
```
MaximilianMeister: `FileUtils.mkdir_p config_dir unless config_dir.directory?`

scripts/barclamp_mgmt_lib.rb
```diff
@@ -67,6 +67,13 @@ def fatal(msg, log = nil, exit_code = 1)
   exit exit_code
 end
 
+def get_yml_paths_from_rpm(component)
+  rpm = "crowbar-#{component}"
+  get_rpm_file_list(rpm).select do |file|
+    file =~ %r!^#{CROWBAR_PATH}/barclamps/([^/]+).yml$!
```
MaximilianMeister: please make sure we don't include all the hound yml files here.
so basically exclude the dot yml files.
they are probably not shipped but for the dev setup this would be necessary

vuntz: This is in the barclamps subdirectory, so won't be an issue.

scripts/barclamp_install.rb
vuntz: Won't work unless I switch the code to Pathname.
(same for all your other comments)

Shall I do that?

scripts/barclamp_mgmt_lib.rb
MaximilianMeister: the issue will be here https://github.com/vuntz/crowbar/blob/barclamp-install-fixes/scripts/barclamp_mgmt_lib.rb#L94

scripts/barclamp_install.rb
MaximilianMeister: right, I would suggest to switch to Pathname

scripts/barclamp_mgmt_lib.rb
vuntz: You misunderstood me: the hound yml files are not there, they're in the parent directory.

vuntz:
@MaximilianMeister all comments should be addressed

scripts/barclamp_mgmt_lib.rb
```diff
@@ -293,24 +292,13 @@ def bc_cloner(item, bc, entity, source, target, replace)
...
-      File.open(new_source, 'r') do |f|
-        s = f.read
-        t.write(bc_replacer(s, bc, entity))
-      end
-      t.close
-      files << new_file
-    end
+    debug "\t\tcopying file #{new_file}."
+    FileUtils.cp new_source, new_file
+    files << new_file
```
rsalevsky: Minor nitpick but I prefer `files.push(new_file)`.

scripts/barclamp_install.rb
```diff
@@ -58,19 +55,41 @@ def usage()
 FileUtils.mkdir log_path unless File.directory? log_path
 log = File.join log_path, "component_install.log"
 
+install_lib_path = Pathname.new("/var/lib/crowbar/install")
+installation_steps_path = install_lib_path.join("installation_steps")
+if installation_steps_path.exist?
+  steps = installation_steps_path.exist?.map do |step|
```
MaximilianMeister: this will not work... you are mapping on a boolean here. :) probably you wanted
`installation_steps_path.readlines.map`

vuntz: Doh, thanks for catching this. Fixed.

scripts/barclamp_install.rb
```diff
@@ -58,19 +55,41 @@ def usage()
...
+end
+crowbar_installed = install_lib_path.join("crowbar-installed-ok").exist?
+
+do_chef = crowbar_installed_barclamps || crowbar_installed
+
 barclamp_yml_files = Array.new
 component_paths = Array.new
 ARGV.each do |src|
   debug "src: #{src}"
+  component_dir = nil
```
tboerger: I think you don't need to define it.

MaximilianMeister: will it then not throw an exception in line 85?

scripts/barclamp_mgmt_lib.rb
vuntz: done

scripts/barclamp_install.rb
vuntz: @tboerger better?

vuntz:
cc @tboerger @MaximilianMeister @rsalevsky 

FWIW, this is needed for the branding subpackage split, which is something that the storage team needs.

rsalevsky:
@vuntz Gating fails while installing crowbar.

vuntz:
> @vuntz Gating fails during installing crowbar.

See my message in the PR:

> Note that gating cannot be green as this requires some changes in the packaging (the yml files and manifests need to be put directly in the right location by the packages, which makes sense anyway)

rsalevsky:
Sorry missed that. +1

toabctl:
+1

PR #2196, opened 2015-12-03
install-chef-suse: Do not stop apache nor crowbar during install

vuntz:
The new webui installer obviously requires these services to stay
running!

cc @MaximilianMeister


MaximilianMeister:
:+1:
tested manually, no issues. lets see what gating says

jsuchome:
+1 (not tested, but makes sense)

aplanas:
+1

tboerger:
:+1: but gating is red.

PR #2198, opened 2015-12-11
Adopt install-chef-suse.sh script for openSUSE Leap

mjura:


aspiers:
Some indentation is breaking Travis.

scripts/install-chef-suse.sh
```diff
@@ -575,9 +584,12 @@ if [ -n "$CROWBAR_FROM_GIT" ]; then
         add_ibs_repo http://dist.suse.de/ibs/Devel:/Cloud:/6/SLE_12_SP1/ cloud
     fi
 
+    # Be sure that systemd-logger is deinstalled
+    zypper --non-interactive remove systemd-logger
```
tboerger: Why do you uninstall it?

dirkmueller:
do not merge before M8 is released

scripts/install-chef-suse.sh
dirkmueller: that looks like a workaround for https://github.com/crowbar/crowbar-core/pull/160

scripts/install-chef-suse.sh
```diff
@@ -575,9 +584,12 @@ if [ -n "$CROWBAR_FROM_GIT" ]; then
...
+
     # install chef and its dependencies
     zypper -n --gpg-auto-import-keys in ruby2.1-rubygem-chef-server ruby2.1-rubygem-chef \
-        rabbitmq-server couchdb ruby2.1-rubygem-activesupport
+        rabbitmq-server couchdb sqlite3 ruby2.1-rubygem-activesupport
```
dirkmueller: why is sqlite3 needed explicitely?

scripts/install-chef-suse.sh
tboerger: Yes, got an explanation why we need that.

scripts/install-chef-suse.sh
tboerger: Maybe because we don't have a pattern yet and this is required by the rails app? Just my guess.

jdsn:
**NOTE**
I am closing this PR because due to a bug in our automation the last commit in the PR collected 1000 status changes. This is the limit for status reports per sha1 (commit).
So feel free to reopen this PR **after** pushing a new commit onto this PR (or forcepushing a small amended change to the existing commits).

If you have questions, ask @jdsn.

PR #2199, opened 2015-12-14
use ntpq to detect live ntp servers (boo#958721)

bmwiedemann:
https://bugzilla.opensuse.org/show_bug.cgi?id=958721#c2
explains that ntpdc is now disabled
so we cannot use it anymore for detecting ntp servers from ntpd

### DIFF
--- a/scripts/install-chef-suse.sh
+++ b/scripts/install-chef-suse.sh
@@ -894,7 +894,7 @@ if [ -z "$custom_ntp_servers" ]; then
     ntp_servers="$( awk '/^server / && ! / 127\./ { print $2 }' < /etc/ntp.conf )"
     if [ -z "$ntp_servers" ]; then
         echo "Auto-detecting existing ntp servers from ntpd ..."
-        ntp_servers="$( ntpdc -l | awk '/client/ && ! /LOCAL/ { print $2 }' )"
+        ntp_servers="$( ntpq -wp | tail -n +3 | cut -c2- | grep '^[^ ]' | awk '! /LOCL/ { print $1 }' )"
     fi
     if [ -n "$ntp_servers" ]; then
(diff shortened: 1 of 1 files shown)


rsalevsky:
+1

aplanas:
-1

`ntpq` truncate the output, cutting the name of the remote.  Use `ntpq -wp`, but the awk needs to be updated too

bmwiedemann:
ntpq: illegal option -- w
:-(

bmwiedemann:
updated to ignore the wrapped lines that start with more than 1 space

aplanas:
+1

aplanas:
tested: +1

scripts/install-chef-suse.sh
```diff
@@ -894,7 +894,7 @@ if [ -z "$custom_ntp_servers" ]; then
     ntp_servers="$( awk '/^server / && ! / 127\./ { print $2 }' < /etc/ntp.conf )"
     if [ -z "$ntp_servers" ]; then
         echo "Auto-detecting existing ntp servers from ntpd ..."
-        ntp_servers="$( ntpdc -l | awk '/client/ && ! /LOCAL/ { print $2 }' )"
+        ntp_servers="$( ntpq -wp | tail -n +3 | grep '^.[^ ]' | awk '! /LOCL/ { print $1 }' | sed 's/[+*]//' )"
```
jdsn: also minus can be in there:
```sed 's/[+-*]//'```

bmwiedemann:
rewritten to use cut to remove the first char in line. also makes the grep easier.

aplanas:
still +1

jdsn:
:+1:

tboerger:
:+1:

PR #2200, opened 2015-12-14
Add branding_install.rb script

vuntz:
This is a small utility to regenerate files that need to be updated once
Crowbar branding package has been installed.

This is needed for the branding subpackage; this could however live directly as a source file in our crowbar package. What do people think?

### DIFF
--- a/scripts/branding_install.rb
+++ b/scripts/branding_install.rb
@@ -1,0 +1,52 @@
+#!/usr/bin/env ruby
+#
+# Copyright 2015, SUSE LINUX Products GmbH
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#  http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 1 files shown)


tboerger:
First of all you should remove the `.rb` extensions. After that you should not rely on `barclamp_mgmt_lib` as this will be dropped with next version of crowbar ;)

vuntz:
> First of all you should remove the .rb extensions.

I'm just being consistent with barclamp_install.rb. Yes, it's horrible, but I'm in favor of consistency. I'm fine renaming all the utilities; can be done before or after this PR (if before, I'll update this PR).

> After that you should not rely on barclamp_mgmt_lib as this will be dropped with next version of crowbar ;)

Hrm, but the only real important piece of code in the small utility is the call to `generate_assets_manifest`. Do you want to copy the method here instead?

tboerger:
> Hrm, but the only real important piece of code in the small utility is the call to generate_assets_manifest. Do you want to copy the method here instead?

Sorry, forget it. I were talking about `barclamp_lib.rb`.

> I'm just being consistent with barclamp_install.rb. Yes, it's horrible, but I'm in favor of consistency. I'm fine renaming all the utilities; can be done before or after this PR (if before, I'll update this PR).

Than let's do it afterwards for both scripts.

tboerger:
:+1:

aplanas:
Even if is only called in OBS / IBS, for me makes sense to be here.

+1

tboerger:
I think I will build some different commands for barclamp_* soon as well but for now it's fine like it is.

rsalevsky:
+1

PR #2202, opened 2016-01-14
drop rubocop from Code Climate

aspiers:
We already have Hound CI which is a bit nicer than Code Climate for rubocop linting, because it comments on style failures inline within the pull request.  Plus Code Climate's rubocop seems to be triggering some false failures.


vuntz:
+1

rsalevsky:
+1

aplanas:
+1

PR #2203, opened 2016-01-15
install-chef-suse: verify tables exist before deleting on failed install

tserong:
If initial installation fails before production.sqlite3 has been created
or populated, subsequent runs of the install script will bail out early
with "Error: no such table: proposals" when it tries to do some cleanup.
This change checks to make sure those tables actually exist before
attempting to delete them.

### DIFF
--- a/scripts/install-chef-suse.sh
+++ b/scripts/install-chef-suse.sh
@@ -327,7 +327,13 @@ fi
 if [ -f $crowbar_install_dir/crowbar-install-failed ] || [ "$CROWBAR_WIZARD_MODE" ]; then
     rm -f $crowbar_install_dir/crowbar-install-failed
     rm -f $installation_steps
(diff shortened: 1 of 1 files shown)


vuntz:
Hrm, I think the intent was to replace this code with a call to what @tboerger added in https://github.com/crowbar/crowbar-core/pull/141

@tboerger what's the status on that?

tboerger:
> @tboerger what's the status on that?

I have no idea, that was a task for @rsalevsky

tboerger:
@rsalevsky feel free to integrate https://github.com/tboerger/crowbar-core/commit/2a03287e0dbc7773e42a9d46dc45232bebf5c771

rsalevsky:
Closing in favor of https://github.com/crowbar/crowbar/pull/2204

PR #2205, opened 2016-01-27
use https links

bmwiedemann:
whereever possible
and fix houndci URL

### DIFF
--- a/.hound.suse.yml
+++ b/.hound.suse.yml
@@ -2,7 +2,7 @@
 # at https://github.com/SUSE/style-guides/blob/master/rubocop-suse.yml
 #
 # The configuration is tested and used with the Rubocop version used by Hound
-# (http://hound-ci.com).
+# (https://houndci.com).
 #
 # We use the default Hound config as a baseline:
 #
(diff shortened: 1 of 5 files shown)


Gemfile
```diff
@@ -6,7 +6,7 @@
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
 #
-#   http://www.apache.org/licenses/LICENSE-2.0
+#   https://www.apache.org/licenses/LICENSE-2.0
```
vuntz: Please no: look at the content at the URL and see the end, the recommendation is to use the http URL. Changing this to https might break automatic license detection.

bmwiedemann:
updated

scripts/barclamp_mgmt_lib.rb
```diff
@@ -229,7 +229,7 @@ def generate_navigation
     out.puts '# you may not use this file except in compliance with the License.'
     out.puts '# You may obtain a copy of the License at'
     out.puts '#'
-    out.puts '#   http://www.apache.org/licenses/LICENSE-2.0'
+    out.puts '#   https://www.apache.org/licenses/LICENSE-2.0'
```
vuntz: You forgot that one :-)

vuntz:
+1 after fixing the last comment.

rsalevsky:
+1

PR #2206, opened 2016-02-09
install-chef-suse: Use same dialog size and text before/in spinner

vuntz:
It seems the --keep-tite option in dialog is creating some flashing for
me now (not sure why), which makes it very visible that the dialog was
different before the spinner and during the spinner.

With this, the flash still occurs, but is actually bearable.


aplanas:
+1

rsalevsky:
+1

PR #2207, opened 2016-02-10
install-chef-suse: Do not call rake as root

vuntz:
When we do this, it creates /var/log/crowbar/production.log owned by
root, which makes the rails app unstartable later on because it can't
write to the log.


aplanas:
+1

jsuchome:
+1

PR #2208, opened 2016-02-11
unset http_proxy variable for knife (bsc#958716)

rsalevsky:


vuntz:
This is unsetting it for the whole install process afterwards. Is this wanted? If yes, why not do it at the beginning? If no, then I guess you need to save the old value and set it again :-)

rsalevsky:
I need to check if this is really this final solution, I just took this fix from the bugreport. Maybe @dirkmueller can give more information?

scripts/install-chef-suse.sh
```diff
@@ -665,7 +665,16 @@ echo_summary "Performing initial chef-client run"
 service chef-client status &> /dev/null && service chef-client stop
 
 if ! [ -e ~/.chef/knife.rb -a -e ~/.chef/root.pem ]; then
-    yes '' | knife configure -i
+    # no_proxy is currently not suppoted in ruby see bsc#958716
+    # unset it for this call
+    if [[ -n $http_proxy ]]; then
+        local proxy = $http_proxy
```
jdsn: syntax error, this must read:
```
local proxy=$http_proxy
```

jdsn:
My proposal without code duplication, and a little more readable (IMHO):
```
proxy_save=$http_proxy
unset $http_proxy
yes '' | knife configure -i
if [[ $proxy_save ]] ; then
  http_proxy=$proxy_save
fi
```

scripts/install-chef-suse.sh
```diff
@@ -665,7 +665,14 @@ echo_summary "Performing initial chef-client run"
 service chef-client status &> /dev/null && service chef-client stop
 
 if ! [ -e ~/.chef/knife.rb -a -e ~/.chef/root.pem ]; then
+    # no_proxy is currently not suppoted in ruby see bsc#958716
+    # unset it for this call
+    proxy_save=$http_proxy
+    unset http_proxy
     yes '' | knife configure -i
+    if [[ $proxy_save ]] ; then
+      http_proxy=$proxy_save
```
bmwiedemann: travis complains about this line: Indent not multiple of 4

bmwiedemann:
one error: unset http_proxy needs to be written without $
and then we could probably avoid the backup+restore of the value by using a subshell with
```
( unset http_proxy ; yes '' | knife configure -i )
```
or even without a sub-shell
```
yes '' | http_proxy= knife configure -
```

rsalevsky:
@bmwiedemann done

scripts/install-chef-suse.sh
```diff
@@ -665,7 +665,9 @@ echo_summary "Performing initial chef-client run"
 service chef-client status &> /dev/null && service chef-client stop
 
 if ! [ -e ~/.chef/knife.rb -a -e ~/.chef/root.pem ]; then
-    yes '' | knife configure -i
+    # no_proxy is currently not suppoted in ruby see bsc#958716
+    # unset it for this call
+    (unset http_proxy; yes '' | knife configure -i)
```
jdsn: install-chef-suse.sh runs with `set -e`
Moving the knife call to a subshell would prevent the -e to exit install-chef-suse.sh if knife exits with an error
So $? needs to be evaluated right after the subshell and exit if it is != 0

jdsn:
:+1:

scripts/install-chef-suse.sh
```diff
@@ -665,7 +665,11 @@ echo_summary "Performing initial chef-client run"
...
+    ret=$?
+    [[ $ret != 0 ]] && exit $ret
```
bmwiedemann: could also be done with  || exit $?

jdsn: agreed

bmwiedemann:
+1

scripts/install-chef-suse.sh
```diff
@@ -665,7 +665,9 @@ echo_summary "Performing initial chef-client run"
 service chef-client status &> /dev/null && service chef-client stop
 
 if ! [ -e ~/.chef/knife.rb -a -e ~/.chef/root.pem ]; then
-    yes '' | knife configure -i
+    # no_proxy is currently not suppoted in ruby see bsc#958716
```
vuntz: typo: supported

vuntz:
+1

mjura:
+1

vuntz:
I think we're missing the backport to 3.0 for that one, aren't we?

rsalevsky:
@vuntz see #2214

vuntz:
> @vuntz see #2214

That's a backport to tex, not 3.0.

rsalevsky:
@vuntz Ups, created a new PR see #2216

PR #2209, opened 2016-02-12
drop unused if clause and small refactor around createrepo

jdsn:
_description updated_

The if condition is always true when not using quotes, thus the else part is never run.
The remaining is just small refactor around createrepo, nice error message and code shortening.

### DIFF
--- a/scripts/install-chef-suse.sh
+++ b/scripts/install-chef-suse.sh
@@ -418,7 +418,9 @@ check_or_create_ptf_repository () {
         if ! [ -e "/srv/tftpboot/suse-$version/$arch/repos/$repo/repodata/repomd.xml" ]; then
             echo "Creating repo skeleton to make AutoYaST happy."
             mkdir -p /srv/tftpboot/suse-$version/$arch/repos/$repo
-            /usr/bin/createrepo /srv/tftpboot/suse-$version/$arch/repos/$repo
+            if ! /usr/bin/createrepo /srv/tftpboot/suse-$version/$arch/repos/$repo ; then
+                die "Createrepo failed to create a repository in /srv/tftpboot/suse-$version/$arch/repos/$repo"
+            fi
         fi
     fi
 }
--- a/scripts/install-chef-suse.sh
+++ b/scripts/install-chef-suse.sh
@@ -416,18 +416,11 @@ check_or_create_ptf_repository () {
         echo "Skipping check for $repo ($version) due to \$REPOS_SKIP_CHECKS"
     else
         if ! [ -e "/srv/tftpboot/suse-$version/$arch/repos/$repo/repodata/repomd.xml" ]; then
-            # Only do this for CROWBAR_FROM_GIT, as usually the crowbar package
-            # creates the repo metadata for PTF
-            if [ -n $CROWBAR_FROM_GIT ]; then
-                echo "Creating repo skeleton to make AutoYaST happy."
-                if ! [ -d /srv/tftpboot/suse-$version/$arch/repos/$repo ]; then
-                    mkdir /srv/tftpboot/suse-$version/$arch/repos/$repo
-                fi
-                /usr/bin/createrepo /srv/tftpboot/suse-$version/$arch/repos/$repo
-            else
-                die "$repo ($version / $arch) has not been set up correctly; \
-                    did the crowbar rpm fail to install correctly?"
+            echo "Creating repo skeleton to make AutoYaST happy."
+            if ! [ -d /srv/tftpboot/suse-$version/$arch/repos/$repo ]; then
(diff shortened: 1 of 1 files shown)


jsuchome:
+1

aplanas:
+1

vuntz:
> The call to createrepo has no error handling. Errors during createrepo are hidden as the installation fails at a later point in time.

This won't be hidden, as install-suse-cloud aborts on any failed command. But adding a nice message with die is still useful :-)

vuntz:
Can someone test this without the mkcloud magic that creates all repos? I'm wondering if we're really safe with the change now.

scripts/install-chef-suse.sh
```diff
@@ -418,12 +418,14 @@ check_or_create_ptf_repository () {
...
-            if [ -n $CROWBAR_FROM_GIT ]; then
+            if [ -n "$CROWBAR_FROM_GIT" ]; then
                 echo "Creating repo skeleton to make AutoYaST happy."
                 if ! [ -d /srv/tftpboot/suse-$version/$arch/repos/$repo ]; then
                     mkdir /srv/tftpboot/suse-$version/$arch/repos/$repo
                 fi
-                /usr/bin/createrepo /srv/tftpboot/suse-$version/$arch/repos/$repo
+                if ! /usr/bin/createrepo /srv/tftpboot/suse-$version/$arch/repos/$repo ; then
+                    die "Createrepo failed to create a repository in /srv/tftpboot/suse-$version/$arch/repos/$repo"
+                fi
```
vuntz: To clarify my concern: I'm thinking that we don't have the repomd.xml file in the packages, so if we don't enter this `if` block, we go in the `else` block and die...

TBH, I'm not sure we really need the `if [ -n "$CROWBAR_FROM_GIT" ]; then`...

jdsn:
> Can someone test this without the mkcloud magic that creates all repos? I'm wondering if we're really safe with the change now.

I use this patch in my manual mkcloud runs already. They fail on different places though.

jdsn:
also tracked as https://bugzilla.suse.com/show_bug.cgi?id=966453

vuntz:
> I use this patch in my manual mkcloud runs already.

What is "manual mkcloud"? Just to clarify my request (to see if it matches what you did): I'd like this to be tested without any of the helpers function that mkcloud is using as mkcloud is creating repos and so on. So that's not the typical setup.

jdsn:
> What is "manual mkcloud"?

I wante to say that I use this patch for cloud deployments that I did manually but still using mkcloud.

> I'd like this to be tested without any of the helpers function that mkcloud is using as mkcloud is creating repos and so on.

That will be reported by the gating job of this PR.

scripts/install-chef-suse.sh
jdsn: > we don't have the repomd.xml file in the packages, so if we don't enter this `if` block

But it is negated
```
if ! [ -e "/srv/tftpboot/suse-$version/$arch/repos/$repo/repodata/repomd.xml" ]
```
so the if block will always be entered.

jdsn: > TBH, I'm not sure we really need the `if [ -n "$CROWBAR_FROM_GIT" ]; then`

That came to my mind as well, because it wasn't an issue until now.
I just found these bugs while debugging something else.

vuntz:
> That won't be reported by the gating job of this PR. So we need a manual test, right?

Yes. But maybe it's not that important for now -- I still fail to understand what is broken without your patch (except that we're more tolerant than we might want to be, but that should be okay).

jdsn:
> I still fail to understand what is broken without your patch

I understand, and meanwhile I agree that it is not needed.

This PR came from a debugging session of a different issue and it looked like being the cause, as the error was logged but the installation continued and failed later.
What we found out later was, that the real cause for this was two install-suse-cloud scripts running in parallel: https://github.com/crowbar/crowbar-core/pull/318

So this PR is just a byproduct of this and not urgend - sorry for the noise. I hope this clarifies it.

vuntz:
Thanks for the clarification; let's look at it post-release.

dirkmueller:
please add the explanation to the git commit message, not to the PR description..

jdsn:
> please add the explanation to the git commit message, not to the PR description..

The commit messages already contain the explanation :)
But as we found out the fix in this PR is wrong. The workflow relied on the if statement being true always. So the bug was a feature.

I will rework this PR.

vuntz:
Not sure we need an error message for createrepo, but why not... +1

jdsn:
> Not sure we need an error message for createrepo, but why not... +1

The intended change is not the nice error message, but rather to die on error. In that case a nice error message makes sense in order to give a hint about the failure.

rsalevsky:
+1

PR #2210, opened 2016-02-15
install-chef-suse: Remove setting REPOS_SKIP_CHECKS env var for DEPS iso

vuntz:
We don't check these repos in this script anymore.


aplanas:
There are some references here:

https://github.com/vuntz/crowbar/blob/no-repos-skip/scripts/install-chef-suse.sh#L416

vuntz:
> There are some references here:
> 
> https://github.com/vuntz/crowbar/blob/no-repos-skip/scripts/install-chef-suse.sh#L416

But they're only for PTF repos, not for Pool/Updates. Maybe the commit message needs to be clearer?

aplanas:
Ah OK, is only dropping the extension of this in the media

aplanas:
+1

mjura:
+1

PR #2211, opened 2016-02-17
Create manifests dir before trying to use it

rsalevsky:
I had this issue in my dev-setup.

### DIFF
--- a/scripts/barclamp_install.rb
+++ b/scripts/barclamp_install.rb
@@ -147,7 +147,7 @@ if do_chef
 end
 
 generate_navigation
-generate_assets_manifest
+generate_assets_manifest if from_rpm
 catalog
 
 exit exit_code


tboerger:
:+1:

scripts/barclamp_mgmt_lib.rb
```diff
@@ -260,6 +260,8 @@ def generate_assets_manifest
 
   manifests = Pathname.new(CROWBAR_PATH).join("barclamps", "manifests")
 
+  manifests.mkdir unless manifests.exist?
```
vuntz: Hrm, shouldn't we just return if the directory doesn't exist? Why create it?

rsalevsky: Then you need to create it manual, don't know if this makes sense.

vuntz: Why? Where is it needed outside this function?

rsalevsky: No, what I mean when I run the barclamp_install.rb in my dev setup the first time (the folder doesn't exist) the it fail due to the manifest can not written. I need to create the folder by hand before running the script then it works.

This is just the solution which I use since some weeks...

vuntz: Which manifest file cannot be written? We're just reading files from the directory here (and I suggested to skip that step if there's no directory). Or do you mean the one in /public/assets/?

rsalevsky: Oh sorry I mean read. https://github.com/rsalevsky/crowbar/blob/create_dir/scripts/barclamp_mgmt_lib.rb#L267 fails

tboerger: I also run into issues within the development environment without that fix.

vuntz: > Oh sorry I mean read. https://github.com/rsalevsky/crowbar/blob/create_dir/scripts/barclamp_mgmt_lib.rb#L267 fails

So we're back to my first comment, why not simply return instead of doing nothing in the method?

rsalevsky: Do you mean return `true` or `false`?

vuntz: Neither, I mean returning from `generate_assets_manifest` (which doesn't return any value).

tboerger: IMHO it doesn't hurt to make sure this directory simply exists.

rsalevsky: > IMHO it doesn't hurt to make sure this directory simply exists.

Same opinion, we can't sure if this dir isn't missing somewhere else.

vuntz: I can tell you it isn't used for anything else -- it's only a hack used for rpm packages...

I don't mind creating the directory that much. It's just not what I would have done -- and the small thread we have here is just because we were all confused about why it's needed (while it's not, it's just one of the two simple ways to fix it).

MaximilianMeister: I have set up a development environment today (after not using it for ~ a month), and I haven't run into this error somehow.... when should this occur?

rsalevsky:
@vuntz I changed the implementation now completely.

scripts/barclamp_install.rb
```diff
@@ -147,7 +147,7 @@ def usage()
 end
 
 generate_navigation
-generate_assets_manifest
+generate_assets_manifes if from_rpm
```
vuntz: Typo: you lost the t :-)

tboerger:
:+1:

MaximilianMeister:
+1

mjura:
+1

PR #2212, opened 2016-02-19
sledgehammer: Handle hardware booting from interface other than eth0

vuntz:
The SLE12 sledgehammer image is hard-coding eth0 as the interface to use
(in /etc/sysconfig/network). But if the node booted from another
interface, we need to have it configured before going on.

### DIFF
--- a/sledgehammer/start-up.sh
+++ b/sledgehammer/start-up.sh
@@ -100,6 +100,11 @@ if ! suse_ver 12; then
     # Make sure our PXE interface is up, then fire up DHCP on it.
     ip link set "$BOOTDEV" up
     dhclient "$BOOTDEV"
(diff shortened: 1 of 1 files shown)


dirkmueller:
-1, doesn't work, we need to rename the interface instead to eth0

vuntz:
> we need to rename the interface instead to eth0

Actually, we went further with Cameron, and it seems a fix like https://github.com/crowbar/crowbar-core/pull/364 is what is missing after this fix.

dirkmueller:
@vuntz okay, I take your word for it, it felt like we still need the rename though. +1 then.

dirkmueller:
merging since its already in the GMC2

PR #2213, opened 2016-02-23
Call knife configure with fixed values (bsc#967811)

rsalevsky:


vuntz:
Can you explain why in the commit message?

scripts/install-chef-suse.sh
```diff
@@ -663,7 +663,23 @@ echo_summary "Performing initial chef-client run"
...
-    yes '' | knife configure -i
+    if [ knife client list | grep -q root ]; then
+        yes "Y" | knife client delete root
+    fi
+    /usr/bin/expect <<EOD
+spawn knife configure -i
+expect {
+  "Overwrite /root/.chef/knife.rb? (Y/N)" { send "Y\r"; exp_continue }
+  "Where should I put the config file? *" { send "/root/.chef/knife.rb\r"; exp_continue }
+  "Please enter the chef server URL: *" { send "\r"; exp_continue }
```
dirkmueller: this needs to be really the localhost url, the one after provisioning

dirkmueller:
-1, the hostname needs to be hardcoded to lcoalhost (or local ip or admin ip). and rationale needs to be given in commit message (to avoid the guesswork code in chef that finds the wrong chef server url).

I think you could also just echo that one parameter to the file if it doesn't exist and then run configure -i, hopefully it will preserve the already set values. (haven't tried that)

jdsn:
+1 (nitpicking: commit message s/then/than/)

rsalevsky:
@vuntz I added a new commit to the PR to make sure the env is correct. But I would like to keep the first commit due to I hit some other issues while testing. For example when `~/.chef/root.pem` is missing then knife asks ` Overwrite /root/.chef/knife.rb? (Y/N)` which is not handled by `yes '' `. Also I hit in the second run that some default values are empty.

mjura:
+1

vuntz:
Needs to be rebased.

scripts/install-chef-suse.sh
```diff
@@ -665,7 +671,25 @@ service chef-client status &> /dev/null && service chef-client stop
 if ! [ -e ~/.chef/knife.rb -a -e ~/.chef/root.pem ]; then
     # no_proxy is currently not supported in ruby see bsc#958716
     # unset it for this call
-    (unset http_proxy; yes '' | knife configure -i)
+    (unset http_proxy
+    if [ knife client list | grep -q root ]; then
+        yes "Y" | knife client delete root
```
vuntz: `knife client delete --yes root` ?

scripts/install-chef-suse.sh
```diff
@@ -665,7 +671,25 @@ service chef-client status &> /dev/null && service chef-client stop
 if ! [ -e ~/.chef/knife.rb -a -e ~/.chef/root.pem ]; then
     # no_proxy is currently not supported in ruby see bsc#958716
     # unset it for this call
-    (unset http_proxy; yes '' | knife configure -i)
+    (unset http_proxy
+    if [ knife client list | grep -q root ]; then
```
vuntz: `grep -q "^ *root$"`

scripts/install-chef-suse.sh
```diff
@@ -665,7 +671,25 @@ service chef-client status &> /dev/null && service chef-client stop
...
+  "Where should I put the config file? *" { send "/root/.chef/knife.rb\r"; exp_continue }
+  "Please enter the chef server URL: *" { send "http://localhost:4000\r"; exp_continue }
+  "Please enter a clientname for the new client: *" { send "root\r"; exp_continue }
+  "Please enter the existing admin clientname: *" { send "chef-webui\r"; exp_continue }
+  "Please enter the location of the existing admin client's private key: *" { send "/etc/chef/webui.pem\r"; exp_continue }
+  "Please enter the validation clientname: *" { send "chef-validator\r"; exp_continue }
+  "Please enter the location of the validation key: *" { send "/etc/chef/validation.pem\r"; exp_continue }
+  "Please enter the path to a chef repository (or leave blank):" { send "\r"; exp_continue }
+}
+EOD
```
vuntz: I'm also wondering if we really need expect here; did you try simply using parameters for knife configure? Something like `knife configure -i -u root --defaults --repository ""`

scripts/install-chef-suse.sh
```diff
@@ -665,7 +671,21 @@ service chef-client status &> /dev/null && service chef-client stop
...
+    fi
+    knife configure \
+    --initial \
+    --yes \
+    --defaults \
+    --user root \
+    --admin-client-name chef-webui \
+    --admin-client-key /etc/chef/webui.pem \
+    --validation-client-name chef-validator \
+    --validation-key /etc/chef/validation.pem \
```
vuntz: These 4 lines are not strictly needed, but okay.

vuntz:
Thanks for the change, +1

jdsn:
:+1: lgtm

PR #2218, opened 2016-04-07
install-chef-suse: Skip detection of ntp servers when no ntp.conf exists

vuntz:
Right now, we are failing because we try to parse ntp.conf. We know
there's no ntp running anyway if there's no file, so we can skip the
detection completely.


toabctl:
+1

rsalevsky:
+1

aplanas:
+1

PR #2219, opened 2016-04-07
install-chef-suse: Fix usage of DEPS iso

vuntz:
The test was wrong, and the Cloud repo is gone now (integrated in the
DEPS iso).


toabctl:
+1

dirkmueller:
+1

PR #2220, opened 2016-04-13
Replace old chef uploader code with new rake task

rsalevsky:


tboerger:
Awesome :+1:

scripts/barclamp_mgmt_lib.rb
```diff
@@ -421,40 +421,10 @@ def bc_install_layout_1_app(from_rpm, bc_path)
...
-    upload_cookbooks_from_dir cookbooks, ['ALL'], log
-    upload_data_bags_from_dir databags, log
-    upload_roles_from_dir roles, log
-  end
-
-  puts "Chef components for (#{components.join(", ")}) (format v1) uploaded."
+# sync the chef parts
+def bc_install_layout_1_chef
+  rails_env = ENV["RAILS_ENV"] || "production"
+  system "cd #{CROWBAR_PATH} && RAILS_ENV=#{rails_env} bin/rake chef:upload:all"
```
MaximilianMeister: why not just

```ruby
system "RAILS_ENV=#{rails_env} #{CROWBAR_PATH}/bin/rake chef:upload:all"
```

MaximilianMeister:
+1

scripts/barclamp_mgmt_lib.rb
vuntz: Please use `su`to not run this as root. We keep having issues with rake being called as root and creating log files that crowbar then cannot write to.

vuntz: (see a couple of lines below for how it's done for another rake call)

rsalevsky:
@MaximilianMeister @tboerger Can you please review again? I made some changes after your review.

MaximilianMeister:
I am not sure I understand the second commit, as chef-client is running before that code that you have added

scripts/install-chef-suse.sh
```diff
@@ -708,6 +708,17 @@ EOF
 
 $chef_client
 
+if ! [ -e /opt/dell/crowbar_framework/config/client.pem ]; then
+    (export EDITOR=/bin/true
```
MaximilianMeister: do we need an `export` here for just this command?

tboerger:
:+1:

jdsn:
:+1:

PR #2223, opened 2016-04-20
Abort installation if system doesn't have FQDN (bsc#974636)

tserong:
Various things break in strange ways if the Crowbar admin node doesn't
have a fully qualified domain name.

### DIFF
--- a/scripts/install-chef-suse.sh
+++ b/scripts/install-chef-suse.sh
@@ -352,6 +352,11 @@ if [ -n "$SSH_CONNECTION" -a -z "$STY" ] && [ -z "$CROWBAR_WIZARD_MODE" ]; then
     die "Not running in screen. Please use \"screen $0\" to avoid problems during network re-configuration. Aborting."
 fi
 
+# All sorts of odd things break if the system has only a shortname, not an FQDN
+if [ -z "$DOMAIN" ]; then
+    die "System not configured with fully qualified domain name.  Aborting."
+fi
+
 rootpw=$( getent shadow root | cut -d: -f2 )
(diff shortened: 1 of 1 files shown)


tboerger:
:-1: as this is already part of https://github.com/crowbar/crowbar-core/blob/master/crowbar_framework/lib/crowbar/checks/network.rb#L25-L31 and it have been dropped from this script as we want to get rid of it entirely on long term.

MaximilianMeister:
@tserong maybe you just have to rebase your code to https://github.com/crowbar/crowbar-core/pull/386 which has been merged 6 days ago. if there is no fqdn, you will be redirected to a sanity checks page and the other api endpoints are blocked

tserong:
OK, fair enough. I'm fine if this is rejected bearing in mind @tboerger and @MaximilianMeister's comments, but I'm keeping it in my SES 3 crowbar package, because that's based on SOC 6, and does still include the install script.

MaximilianMeister:
@tserong okay,  we have also backported the patch to cloud 6 https://github.com/crowbar/crowbar-core/pull/456

just for reference https://github.com/crowbar/crowbar/pull/2159/ was the PR where the check was dropped. I think in your case if the admin node doesn't have a valid FQDN and you try to restart crowbar, you will get a rails error because this check has been moved to the rails app's initializer. see https://github.com/crowbar/crowbar-core/pull/61

dirkmueller:
@tserong can we close this given the last comments?

tserong:
@dirkmueller yeah, go ahead.

PR #2226, opened 2016-04-29
pin listen version to 3.0.7 to avoid ruby 2.2 dependency

MaximilianMeister:


tboerger:
Maybe we should just upgrade ruby.
Otherwise LGTM

MaximilianMeister:
In the meantime there has been a new minor version 3.0.7

I have just updated it

dirkmueller:
works for me, +1

PR #2227, opened 2016-05-24
Switch to SP2

rsalevsky:


scripts/install-chef-suse.sh
```diff
@@ -522,7 +522,7 @@ for arch in $product_arches; do
         if ! is_ses; then
```
vuntz: Not visible here, but on line 518, you probably need to change SLES12-SP1 to SLES12-SP2 (unless SP2 still has the bug where it pretends to be SP1...)

scripts/install-chef-suse.sh
```diff
@@ -501,7 +501,7 @@ fi
 
 for arch in $product_arches; do
     # Checks for SLE12 SP1 medias
```
vuntz: s/SP1/SP2/

scripts/install-chef-suse.sh
rsalevsky: Fixed

jsuchome:
+1

jsuchome:
I assume this cannot turn green without https://github.com/SUSE-Cloud/automation/pull/995 ...

scripts/install-chef-suse.sh
```diff
@@ -542,10 +542,10 @@ fi
 
 
 for arch in $supported_arches; do
-    check_or_create_ptf_repository 12.1 $arch PTF
+    check_or_create_ptf_repository 12.2 $arch PTF
```
jsuchome: I think we still need PTF for 12.1 as well

jsuchome:
See https://github.com/crowbar/crowbar/pull/2228

vuntz:
Was cherry-picked in https://github.com/crowbar/crowbar/pull/2228

PR #2228, opened 2016-06-02
Support SLES12 SP2, but not require it for Cloud repository

jsuchome:
Based on https://github.com/crowbar/crowbar/pull/2227

I know that last commit is ugly, but do we need more elegant solution?

### DIFF
--- a/scripts/install-chef-suse.sh
+++ b/scripts/install-chef-suse.sh
@@ -542,9 +542,12 @@ fi
 
 
 for arch in $supported_arches; do
+    # We will need support for SP1 (ceph) and SP2 nodes
+    check_or_create_ptf_repository 12.1 $arch PTF
     check_or_create_ptf_repository 12.2 $arch PTF
 
     # Currently we only sign the PTF repository
+    sign_repositories 12.1 $arch PTF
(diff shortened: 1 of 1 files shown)


jsuchome:
Hm, but we would need to comment all checks for SP2 repos...

vuntz:
> Hm, but we would need to comment all checks for SP2 repos...

I was going to point this out :-) Just disable the checks for now in general, I guess.

rsalevsky:
My plan was to use SP2 as default for the admin node and just check the SP2 repo.

jsuchome:
Yes, that might be done later. Current requirement is to enable SP2 nodes to be installed, but not having it a default. That's why @bmwiedemann added an option for that https://github.com/SUSE-Cloud/automation/pull/1008

vuntz:
+1

rsalevsky:
Same is done in https://github.com/crowbar/crowbar/pull/2227, except disabling repo checks for SP2.

jsuchome:
@rsalevsky indeed, and it is mentioned above (I'm cherry-picking your commit). But those other 2 commits are needed as well.

rsalevsky:
+1

PR #2229, opened 2016-06-02
wait for network to be reachable

abel-navarro:
Sometimes during startup net is not ready before we try to reach NFS. This can make the boot sequence fail. We will wait for 1 minute if the admin node is not reachable.

### DIFF
--- a/sledgehammer/start-up.sh
+++ b/sledgehammer/start-up.sh
@@ -149,6 +149,13 @@ echo "# Sledgehammer added to log to the admin node" >> /etc/rsyslog.conf
 echo "*.* @@${ADMIN_IP}" >> /etc/rsyslog.conf
 service $RSYSLOGSERVICE restart
 
+# Sometimes at this point network is not up yet, wait for it
+echo "Waiting for admin server ($ADMIN_IP) to be reachable; will wait up to 60 seconds..."
+ping -c 1 -w 60 $ADMIN_IP > /dev/null || {
+    echo "Admin server ($ADMIN_IP) not reachable."
+    echo "Things will end badly."
+}
+
 # showmount needs a running rpcbind service
 service rpcbind start
 


nicolasbock:
I presume that 1 minute is plenty to fix this issue, right?

+1

vuntz:
Where did you see this? I'm a bit surprised, as the `ifup` call should just by synchronous for this.

abel-navarro:
This problem shows every now and then in our CI. This card covers it: https://trello.com/c/nA7qGKiu

abel-navarro:
The error this PR is supposed to fix is still happening every now and then (https://ci.suse.de/job/openstack-mkcloud/28683). Any chance the PR can be merged?

sledgehammer/start-up.sh
```diff
@@ -149,6 +149,11 @@ echo "# Sledgehammer added to log to the admin node" >> /etc/rsyslog.conf
 echo "*.* @@${ADMIN_IP}" >> /etc/rsyslog.conf
 service $RSYSLOGSERVICE restart
 
+# Sometimes at this point network is not up yet, wait for it
+ping -c 1 -w 60 $ADMIN_IP > /dev/null || {
+    echo "error: $ADMIN_IP not reachable"
```
aspiers: Do you really want to continue at this point, if the IP is still not reachable?

abel-navarro: If I recall correctly from the tests I made, the admin node just needed a bit more time to be reachable hence the 1 minute buffer time. The failure case is not dealt here. I can take a look at that and abort the script all together.

sledgehammer/start-up.sh
```diff
@@ -149,6 +149,11 @@ echo "# Sledgehammer added to log to the admin node" >> /etc/rsyslog.conf
 echo "*.* @@${ADMIN_IP}" >> /etc/rsyslog.conf
 service $RSYSLOGSERVICE restart
 
+# Sometimes at this point network is not up yet, wait for it
+ping -c 1 -w 60 $ADMIN_IP > /dev/null || {
```
vuntz: I don't think you want `ping -c 1 -w 60`; I think you want to repeat 60 times `ping -c 1 -w 1`. It's slightly different as in the current code, the echo request is only sent once, and might be lost.

sledgehammer/start-up.sh
```diff
@@ -149,6 +149,11 @@ echo "# Sledgehammer added to log to the admin node" >> /etc/rsyslog.conf
 echo "*.* @@${ADMIN_IP}" >> /etc/rsyslog.conf
 service $RSYSLOGSERVICE restart
 
+# Sometimes at this point network is not up yet, wait for it
```
vuntz: Can you put something like:
 echo "Waiting for admin server to be reachable; will wait up to 60 seconds..."

sledgehammer/start-up.sh
vuntz: To adopt the same style as in the earlier code, I would do:
```
echo "Admin server not reachable."
echo "Things will end badly."
```

rsalevsky:
Needs a rebase

sledgehammer/start-up.sh
abel-navarro: I verified this and the way it was written seems to be the correct one. With `ping -c 1 -w 60` we are not sending one packet, we are sending packets for 60 seconds and wait a single reply. If a single reply arrives before the deadline (`-w 60`), ping exits successfully. Otherwise one packet per second will be sent up to 60 and then exit with failure.

I have tested this manually and seems to be working as expected.

from ping man:
```
-w deadline
              Specify a timeout, in seconds, before ping exits regardless of how many packets have been sent or received. In this case ping does not stop after
              count packet are sent, it waits either for deadline expire or until count probes are answered or for some error notification from network.
```

abel-navarro:
Addressed requested changes. Rebased.

sledgehammer/start-up.sh
vuntz: Oh, didn't know that: `With deadline option, ping waits for count ECHO_REPLY packets, until the timeout expires.`

vuntz:
No need to wait for mkcloud, as the change needs to be manually pushed to the package. Will do now.

PR #2231, opened 2016-06-08
Do not require deprecated /etc/sysconfig/clock (bsc#983837)

dirkmueller:
As a stop gap before this code is rewritten to use the systemd
methods, just ignore the failure since the whole feature is optional
anyway.


jgrassler:
+1

nicolasbock:
+1

PR #2233, opened 2016-06-21
Added back checks for SLES12SP2 based repositories

jsuchome:
Added back checks for SLES12SP2 based repositories, to be used
when SLES12SP2 is the default. They were temporary removed in
commit bf5a372a4770b0bb4853a7cc842c20c322fb2695

### DIFF
--- a/scripts/install-chef-suse.sh
+++ b/scripts/install-chef-suse.sh
@@ -499,38 +499,34 @@ else
     product_arches="$supported_arches"
 fi
 
-# FIXME:
-# Before we switch to SP2 as default, temporary disable checking SLES and Cloud repos,
-# so we have option to deploy both SP1 and SP2 based clouds
+for arch in $product_arches; do
+    # Checks for SLE12 SP2 medias
+    MEDIA=/srv/tftpboot/suse-12.2/$arch/install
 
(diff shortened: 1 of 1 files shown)


rsalevsky:
@jsuchome Is this still needed?

jsuchome:
Seems like the code currently does not check for SP2 media, and we've switched already. So I'd say yes, this should be probably used...

scripts/install-chef-suse.sh
```diff
@@ -499,38 +499,34 @@ else
...
-#done
+    check_media_links $MEDIA
+
+    if [ -f $MEDIA/content ] && egrep -q "REPOID.*/suse-openstack-cloud-deps/" $MEDIA/content; then
+        echo "Detected SUSE OpenStack Cloud Deps media."
+    else
+        check_media_content \
+            SLES12-SP2 \
+            $MEDIA \
+            #b52c0f2b41a6a10d49cc89edcdc1b13d
```
vuntz: Might be good to take the checksum from the final SP2 image here.

jsuchome:
I've lost my original branch. Continued in https://github.com/crowbar/crowbar/pull/2304

PR #2235, opened 2016-06-22
Update rubocop configs

rsalevsky:


aspiers:
As per [my comment on the corresponding review in `crowbar-core`](https://github.com/crowbar/crowbar-core/pull/506#issuecomment-227902991), I might be OK with this change, but :-1: until a proper explanation and justification is given in the commit message.  Thanks in advance for the extra info ;-)

.hound.ruby.yml
```diff
@@ -1,13 +1,14 @@
 inherit_from: .hound.suse.yml
 
 AllCops:
-  RunRailsCops: true
+  Rails:
+    Enabled: true
```
nicolasbock: What are those changes doing? Should those be put into a separate PR?

.hound.ruby.yml
```diff
@@ -1,13 +1,14 @@
...
   DisplayStyleGuide: true
   DisplayCopNames: true
 
 Style/RedundantReturn:
   Exclude:
-    - "chef/data_bags/crowbar/migrate/*"
+    - "chef/data_bags/crowbar/migrate/**/*"
```
nicolasbock: Same here, separate PR?

.hound.ruby.yml
rsalevsky: Just a syntax change, the old syntax is deprecated.

.hound.ruby.yml
rsalevsky: Same PR but separate commit.

MaximilianMeister:
+1

nicolasbock:
+1

PR #2236, opened 2016-07-02
install-chef-suse: Run barclamp_install.rb only once

vuntz:
We were calling it a second time just for hyperv. Instead, call it with
hyperv the first time. This saves some time in the install process.


toabctl:
+1

rsalevsky:
+1

dirkmueller:
+1

PR #2237, opened 2016-07-03
install-chef-suse: Use chef-client own lock, not crowbar wrapper lock

vuntz:
We have a wrapper script in crowbar for chef-client
(blocking_chef_client.sh) to avoid locking issues in chef-client. This
is obsolete since chef-client has some locking for ages now. So avoid
using anything related to this wrapper.


dirkmueller:
LGTM

rsalevsky:
+1

PR #2242, opened 2016-07-14
Documented GEM_HOME

jgrassler:
Without setting a dedicated Ruby environment via the GEM_HOME
environment, the development environment is likely to break
the regular crowbar setup on the admin node it is installed
to by installing gems to the system paths. This commit adds
instructions on using GEM_HOME to avoid this. That way a cloud 
can be re-used for other purposes that do not involve the crowbar 
dev environment.

Note: mkcloud's devsetup step [does not use GEM_HOME either](https://github.com/SUSE-Cloud/automation/blob/master/scripts/qa_crowbarsetup.sh#L4580) so it would probably be a good idea to adapt that as well.

### DIFF
--- a/doc/development.md
+++ b/doc/development.md
@@ -100,16 +100,29 @@ Before we can start you need to match some prerequirements on your host machine.
 
     1. Change to ```/opt/crowbar/crowbar_framework```.
 
-    2. Run ```bundle install```
+    2. Set a dedicated GEM_HOME for the dev environment to avoid clobbering the
+       admin node's system gems:
 
-    3. Install all barclamps with this snippet
+       ```
+       export GEM_HOME=/tmp/crowbar-dev-gems
+       ```
+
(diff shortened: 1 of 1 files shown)


rsalevsky:
Just execute bundle with the `--path` option, but this documentation is anyway outdated.

jgrassler:
True, but that's not documented either :-)

What else is outdated? I'd rather fix this (or even remove it entirely) than leave stale documentation around for people to get confused...

Itxaka:
LGTM, always up for not installing gems into the system +1

MaximilianMeister:
@jgrassler overall I wouldn't say it is outdated, to me it looks still valid, but it could have been updated after merging https://github.com/SUSE-Cloud/automation/pull/1027

so this change here lgtm
+1

@rsalevsky after that can you update the docs to fit the `devsetup` step?

PR #2245, opened 2016-07-15
temporary workaround for bsc#988961 for M2

MaximilianMeister:
this should be only temporary until
https://bugzilla.suse.com/show_bug.cgi?id=988961 is fixed

for the description of the problem please have a look at the bug


rsalevsky:
Closing in favor for #2247

PR #2246, opened 2016-07-15
Replace md5 checksum on repository keys with fingerprint (bsc#988907)

dirkmueller:
(cherry picked from commit e1c01647bf36d14f7f35c1fee23d7738d34a13a7)

### DIFF
--- a/releases/tex/master/extra/repos-check-suse
+++ b/releases/tex/master/extra/repos-check-suse
@@ -41,15 +41,15 @@ skip_check_for_repo () {
 check_key_file () {
   key_file=$1
 
-  case "`md5sum $key_file | awk '{print $1}'`" in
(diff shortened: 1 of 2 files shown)


aplanas:
+1

releases/tex/master/extra/repos-check-suse
```diff
@@ -41,15 +41,15 @@ skip_check_for_repo () {
 check_key_file () {
   key_file=$1
 
-  case "`md5sum $key_file | awk '{print $1}'`" in
-    a0857768a900ae6bb1e6c8af62460ce3|b2740847428d13d201642ca9cf89bec7)
```
bmwiedemann: Why are two keys here and we replace it with one? Otherwise +1

dirkmueller: no idea

PR #2247, opened 2016-07-15
Make sure that the chef indexer finished (bsc#988961)

rsalevsky:
After deploying the crowbar role, solr takes some time to index the new
role. This change force rebuilds the index to prevent empty searches.

### DIFF
--- a/scripts/install-chef-suse.sh
+++ b/scripts/install-chef-suse.sh
@@ -323,6 +323,19 @@ wait_for_crowbar ()
     fi
 }
 
+wait_for_chef ()
+{
+    local count=0
+    local value=10000
+    while (($count < 60 && $value != 0))
+    do
+        sleep 1
+        value=$(chef-expanderctl queue-depth | grep total | awk -F: '{ print $2 }')
+        echo "Expander Queue Total = $value"
+        count=$(($count + 1))
+    done
+}
+
 if [ -f $crowbar_install_dir/crowbar-installed-ok ]; then
     run_succeeded=already_before
 
@@ -994,6 +1007,9 @@ touch /var/run/crowbar/deploying
 
 wait_for_crowbar
 
+# Need to make sure that we have the indexer/expander finished
+wait_for_chef
+
 # From here, you should probably read along with the equivalent steps in
 # install-chef.sh for comparison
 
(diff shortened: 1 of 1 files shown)


jdsn:
:+1:

toabctl:
+1 (assuming it was tested and works).

Itxaka:
Seems to work ok locally. I noticed that this only works in chef<11 so maybe we should set a note in case we move from chef10 in the future? from chef>=11 we need to use `chef-server-ctl reindex` instead

+1

vuntz:
Asking out loud: would the code from https://github.com/rsalevsky/crowbar/blob/b9da555dfa69766ff89fdb73c7afe85c5e756052/scripts/install-chef-suse.sh#L1037 help?

vmoravec:
`knife index rebuild -y` turns out not to be working for the successful first script run, I succeeded only after putting `sleep 60` on the next line. 
-1

rsalevsky:
@vuntz We can try it out... Do you think we should move the code up our should we have the check twice?

vmoravec:
@vunt @rsalevsky Moving https://github.com/rsalevsky/crowbar/blob/b9da555dfa69766ff89fdb73c7afe85c5e756052/scripts/install-chef-suse.sh#L1037..L1046 up below the `wait_for_crowbar` line helped, yay!

vuntz:
> @vuntz We can try it out... Do you think we should move the code up our should we have the check twice?

I'd probably do it twice, just to be safe.

vmoravec:
@vuntz Second time it always seems to return only a single line: `Expander Queue Total =        0` , the first time the output is more verbose:
```ruby 
Expander Queue Total =        16
Expander Queue Total =        16        
Expander Queue Total =        15    
Expander Queue Total =        15    
Expander Queue Total =        4    
Expander Queue Total =        4    
Expander Queue Total =        2    
Expander Queue Total =        0    
```

rsalevsky:
I think it doesn't hurt us at the moment to check it twice. Long term we should get anyway rid of this check.

PS: I added the patch to the D:C:7:S package in case someone want's to test it.

scripts/install-chef-suse.sh
```diff
@@ -323,6 +323,19 @@ wait_for_crowbar ()
     fi
 }
 
+wait_for_chef ()
+{
+    local count=0
+    local value=10000
+    while (($count < 60 && $value !=0))
```
vuntz: You might want to add a space before `0`. But minor detail.

vuntz:
+1, thanks for the fix!

aplanas:
+1

toabctl:
+1

rsalevsky:
Just added the missing space  before `0`.

PR #2249, opened 2016-07-15
Wait until the crowbar role is successful deployed (bsc#988961)

rsalevsky:
After deploying the crowbar role it takes some time until it is really
available. This is partly related to the known chef indexer issue.

Alternative to #2247

### DIFF
--- a/scripts/install-chef-suse.sh
+++ b/scripts/install-chef-suse.sh
@@ -323,6 +323,21 @@ wait_for_crowbar ()
     fi
 }
 
+wait_for_crowbar_role ()
+{
+    local crowbar_role=0
+    for i in {0..100}; do
+        knife search node "role:crowbar" | grep -q -e Roles:.*crowbar, && {
+            crowbar_role=1
+            break
+        }
+        sleep 2
+    done
+    if [ $crowbar_role -ne 1 ]; then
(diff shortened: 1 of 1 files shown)


aplanas:
extra bsd# in the commit message

scripts/install-chef-suse.sh
```diff
@@ -323,6 +323,21 @@ wait_for_crowbar ()
     fi
 }
 
+wait_for_crowbar_role ()
+{
+    local crowbar_role=0
+    for ((x=1; x<10; x++)); do
```
aplanas: more clear: `for i in {0..9} ; do`

aplanas:
over all, why a loop and sleep is better than https://github.com/crowbar/crowbar/pull/2247?

vmoravec:
This has failed with `sleep 2`, succeeded after I set `sleep 10`. Beside that I found `ERROR: Errno::EPIPE: Broken pipe @ io_write - <STDOUT>` in the log but it had no impact on the script result

rsalevsky:
@vmoravec Changed the time how often it retries, 200 seconds should be enough.

@aplanas The other one does't work always. It's more like trying to find a fix who works always.

rsalevsky:
Closing in favor of #2247

PR #2251, opened 2016-07-18
Fix last change to work in all locales

dirkmueller:
the grep for "fingerprint" only works in C locale, not if you
use a different locale.

(cherry picked from commit daaa69b76065999337cd1344fd0efb2ff56f77d7)


aplanas:
+1

nicolasbock:
+1

PR #2255, opened 2016-07-27
define $DNS_JSON in the same way as $PROVISIONER_JSON

aspiers:
23794e55b started referring to DNS_JSON without ever defining it, so apparently this ability to override DNS settings during install via `/{etc,root}/crowbar/dns.json` has been broken for 3 years and 3 days ...

Another example of why #shellsucks - undefined variables are silently used without any error or even warning.

### DIFF
--- a/scripts/install-chef-suse.sh
+++ b/scripts/install-chef-suse.sh
@@ -391,11 +391,14 @@ json_read () {
     fi | sed "s/^[^=]*=//g"
 }
 
-if [ -f /etc/crowbar/provisioner.json ]; then
-    PROVISIONER_JSON=/etc/crowbar/provisioner.json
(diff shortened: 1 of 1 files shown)


nicolasbock:
+1

dirkmueller:
@aspiers you can make shell bail out on undefined variables by setting "-u".

vuntz:
@aspiers I'll close, feel free to reopen if you disagree

aspiers:
The bug is there, but not for the reason I describe in this PR.  If you look at the commit this PR refers to, you'll see that I wrote the code you linked to, so it's somewhat embarrassing / amusing that I had forgotten about that and misunderstood the source of the bug.  The problem is that `DNS_JSON` gets defined too late, just a few lines after it is used.

So we still need to fix this, but it will be cleaner to submit a new PR.  That said, I've reopened this one as a reminder that we need to submit a new PR.

PR #2258, opened 2016-08-16
Script to start admin server zypper upgrade

jsuchome:
TODO: package the script

### DIFF
--- a/scripts/upgrade_admin_server.sh
+++ b/scripts/upgrade_admin_server.sh
@@ -1,0 +1,54 @@
+#!/bin/bash
+#
+# This script starts the upgrade of admin server to the version of system and/or
+# Cloud product.
+# The script assumes that the repositories are already correctly changed to point
+# to the new product versions. After the package upgrade is done, script initiates
+# a reboot of the machine.
+
+LOGFILE=/var/log/crowbar/admin-server-upgrade.log
+mkdir -p "`dirname "$LOGFILE"`"
+
+# Copy stdout to fd 3
+exec 3>&1
+# Create fd 4 for logfile
+exec 4>> "$LOGFILE"
+# Set fd 1 and 2 to logfile
+exec 1>&4 2>&1
+
+set -x
+
+upgrade_admin_server()
+{
+    # we will need the dump for later migrating it into postgresql
+    pushd /opt/dell/crowbar_framework
+    export RAILS_ENV=production
+    bin/rake db:migrate
+    bin/rake db:dump
+    popd
+
+    # Upgrade the distribution non-interactively
+    zypper --no-color --releasever 12.2 ref -f
+    zypper --no-color --non-interactive dist-upgrade -l --recommends --replacefiles
+    ret=$?
+    if [ $ret != 0 ]; then
+        # In the failed case, crowbar should tell user to check zypper logs,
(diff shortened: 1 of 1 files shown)


rsalevsky:
Can you add `rake db:migrate and `rake db:dump` before the upgrade command? We need this information to update/migrate the database to PostgreSQL.

jsuchome:
@rsalevsky why `db:migrate` when there's still no new package?

rsalevsky:
@jsuchome We need this be sure that we have the newest schema.rb otherwise the restore can fail.

jsuchome:
@rsalevsky I'm not saying we do not need db:migrate call. But:
- when should it be called? Before package upgrade does not seem to make sense, so right before reboot?
- maybe it could be called from some other function, and let this one do only the upgrade related parts. (Unless the rake calls are needed between upgrade and reboot)

rsalevsky:
It should be done before we upgrade the packages otherwise it will fail due to the missing sqlite packages.

scripts/upgrade_admin_server.sh
```diff
@@ -0,0 +1,34 @@
...
+# to the new product versions. After the package upgrade is done, script initiates
+# a reboot of the machine.
+
+
+upgrade_admin_server()
+{
+    # we will need the dump for later migrating it into postgresql
+    pushd /opt/dell/crowbar_framework
+    bin/rake db:migrate
+    bin/rake db:dump
```
MaximilianMeister: do we need to add `RAILS_ENV=production` before these statements?

scripts/upgrade_admin_server.sh
```diff
@@ -0,0 +1,34 @@
...
+    popd
+
+    # Upgrade the distribution non-interactively
+    zypper --releasever 12.2 ref -f
+    zypper --non-interactive dist-upgrade -l --recommends --replacefiles
```
MaximilianMeister: what happens if the commands come back with an error? should we add some tests here?

jsuchome: Good question. We may create admin-server-upgraded-failure or something like that instead of -ok. Question is, if we should skip the reboot - it might have happened after the os upgrade or before...

scripts/upgrade_admin_server.sh
```diff
@@ -0,0 +1,42 @@
...
+    ret=$?
+    if [ $ret != 0 ]; then
+      # In the failed case, crowbar should tell user to check zypper logs,
+      # fix the errors and continue admin server manually,
+      echo "$ret" > /var/lib/crowbar/install/admin-server-upgrade-failed
+      return
+    fi
```
MaximilianMeister: maybe exit $ret here instead of return. and i think it is also important that we can run this script multiple times, but i think it should be ok as it is.

MaximilianMeister:
can you also make that script log to some file?

scripts/upgrade_admin_server.sh
jsuchome: Hm, that exit $ret does not make sense here, it is running in background and script already exited

rsalevsky:
+1

MaximilianMeister:
+1

rsalevsky:
@jsuchome Can you open a PR also to master?

PR #2261, opened 2016-08-19
indicate ongoing upgrade

MaximilianMeister:
this makes it easier for the crowbar api to determine if the upgrade
is running.
also use a variable for the crowbar install directory.

### DIFF
--- a/scripts/upgrade_admin_server.sh
+++ b/scripts/upgrade_admin_server.sh
@@ -20,6 +20,16 @@ set -x
 
 upgrade_admin_server()
 {
+    local installdir=/var/lib/crowbar/install
+
+    if [[ -f $installdir/admin_server_upgrading ]] ; then
+        echo "Exit: Upgrade already running..."
+        exit 1
+    fi
+
+    # Signalize that the upgrade is running
+    touch $installdir/admin_server_upgrading
+
     # we will need the dump for later migrating it into postgresql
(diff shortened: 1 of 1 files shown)


scripts/upgrade_admin_server.sh
```diff
@@ -20,6 +20,16 @@ set -x
 
 upgrade_admin_server()
 {
+    local installdir=/var/lib/crowbar/install
+
+    if [[ -f $installdir/admin_server_upgrading ]] ; then
+        echo "Exit: Upgrade already running..."
+        exit 1
```
jsuchome: As this function is executed in background, we will  not know the exit value. So the check if the upgrade is already running should be done directly from crowbar, I think.

jsuchome:
+1

scripts/upgrade_admin_server.sh
MaximilianMeister: right, noted

MaximilianMeister: https://github.com/crowbar/crowbar-core/pull/604/commits/092b2ae40161a519a722a202e0e3f88804226c8a

jsuchome: Nice

rsalevsky:
+1

jsuchome:
Please backport to 3.0

PR #2263, opened 2016-08-24
Refactor running rake (bsc#993035)

s-t-e-v-e-n-k:
If barclamp_install fails to upload cookbooks, it does not check the
exit status of the bin/rake call, and merrily continues on, finally
exiting with status 0. Refactor the code that calls bin/rake to
no longer make use of system() or su, and instead use spawn(). This
also means that output from bin/rake chef:upload:all is also logged,
it wasn't previously.

Also clean-up the begin/rescue block when running schema migrations,
the function exits if there is an error, and doesn't throw an
exception.

### DIFF
--- a/scripts/barclamp_mgmt_lib.rb
+++ b/scripts/barclamp_mgmt_lib.rb
@@ -26,6 +26,7 @@ require "active_support/all"
 require "pp"
 require "i18n"
 require "pathname"
+require "etc"
 
 if I18n.respond_to? :enforce_available_locales
   I18n.enforce_available_locales = false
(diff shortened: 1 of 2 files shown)


scripts/barclamp_mgmt_lib.rb
```diff
@@ -421,24 +422,36 @@ def bc_install_layout_1_app(from_rpm, bc_path)
...
-def bc_install_layout_1_chef
+def run_rake_task(task_name, log)
+  crowbar_user = Etc.getpwnam("crowbar")
   rails_env = ENV["RAILS_ENV"] || "production"
-  upload_cmd = "cd #{CROWBAR_PATH} && RAILS_ENV=#{rails_env} bin/rake chef:upload:all"
-  system "su -s /bin/sh - crowbar sh -c \"#{upload_cmd}\""
+  debug("spawning bin/rake #{task_name}")
+  rake_process = spawn({"RAILS_ENV" => rails_env},
+                       "bin/rake --silent #{task_name}",
+                       :uid => crowbar_user.uid,
```
rsalevsky: We can't change the uid or gid from a user. The script is not always executed as root but crowbar is allowed to use su. If the script is executed as crowbar user then setting the uid or gid will fail.

rsalevsky:
The bug happens also on C6 but there the upload logic is completely different. On C6 we use the knife bash cmd instead of a rake task.

scripts/barclamp_mgmt_lib.rb
s-t-e-v-e-n-k: Right, thanks, I'll make sure to test that as well. Maybe I'll be lucky and spawn() won't try and set the uid/gid if the user matches, but I doubt it.

s-t-e-v-e-n-k:
Yes, the bug does explicitly call out C6 too, I was planning to fix it -- but wanted to hit master first.

scripts/barclamp_mgmt_lib.rb
rsalevsky: Can you also test what happens if you execute the script as completely different user? I think this will also not work.

s-t-e-v-e-n-k: I've investigated this. The previous code makes use of su -s /bin/sh - crowbar, which run as crowbar prompts for a password, so I'm wondering if you have confused su and sudo. I've been having trouble running barclamp_install as a non-root user, since it keeps wanting to write into /opt/dell.

rsalevsky: Sorry, for the confusion. You are right I mixed su and sudo.

rsalevsky:
+1

nicolasbock:
+1

vuntz:
+1

nkrinner:
+1

PR #2264, opened 2016-08-24
Various minor fixes for upgrade_admin_server.sh

jsuchome:


scripts/upgrade_admin_server.sh
```diff
@@ -36,6 +36,8 @@ upgrade_admin_server()
     bin/rake db:migrate
     bin/rake db:dump
     popd
+    chown crowbar:crowbar /opt/dell/crowbar_framework/db/schema.rb
+    chown crowbar:crowbar /opt/dell/crowbar_framework/db/data.yml
```
MaximilianMeister: why not simply call:
```
sudo -u crowbar bin/rake db:migrate
sudo -u crowbar bin/rake db:dump
```

jsuchome: That's also the option, but the result is the same

rsalevsky: I prefer Maximilian's option.

MaximilianMeister:
thanks for the update, just needs to be backported (a github label like `needs backport` would be nice)

+1

rsalevsky:
+1

PR #2266, opened 2016-08-25
Stop duplicating path settings

s-t-e-v-e-n-k:
In barclamp_mgmt_lib, we check if $CROWBAR_PATH is set and base all
other paths on it. We then duplicate the code almost exactly for
the case when it isn't set. Clean it up.

### DIFF
--- a/scripts/barclamp_mgmt_lib.rb
+++ b/scripts/barclamp_mgmt_lib.rb
@@ -38,21 +38,15 @@ MODEL_SUBSTRING_CAPSS = '==%BC-MODEL=='
 
 if ENV["CROWBAR_DIR"]
   BASE_PATH = ENV["CROWBAR_DIR"]
-  BARCLAMP_PATH = File.join BASE_PATH, 'barclamps'
-  CROWBAR_PATH = File.join BASE_PATH, 'crowbar_framework'
-  MODEL_SOURCE = File.join CROWBAR_PATH, 'barclamp_model'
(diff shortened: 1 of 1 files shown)


toabctl:
+1

aspiers:
+1. Technically it doesn't seem consistent to me to freeze some constants and then omit freezing the ones which hound is too stupid to realise are constants, so you could freeze those too, but I'm not too bothered either way.

scripts/barclamp_mgmt_lib.rb
```diff
@@ -38,21 +38,15 @@
...
-  BASE_PATH = File.join '/opt', 'dell'
-  BARCLAMP_PATH = File.join BASE_PATH, 'barclamps'
-  CROWBAR_PATH = File.join BASE_PATH, 'crowbar_framework'
-  MODEL_SOURCE = File.join CROWBAR_PATH, 'barclamp_model'
-  BIN_PATH = File.join BASE_PATH, 'bin'
-  UPDATE_PATH = '/updates'
-  ROOT_PATH = '/'
+  BASE_PATH = File.join "/opt", "dell"
 end
+BARCLAMP_PATH = File.join BASE_PATH, "barclamps"
```
rsalevsky: I you want to make the refactoring perfect then it would be nice if you can add (). e.g `File.join(BASE_PATH, "barclamps")` hound is not able to detect this.

nicolasbock:
+1

PR #2267, opened 2016-09-07
Fix rubocop ruby version

rsalevsky:
Rubocop should test against the ruby version we use otherwise it can
happen that it shows wrong messages.

For example https://github.com/crowbar/crowbar-openstack/pull/516#discussion_r77812739 which is a ruby 2.3 feature.


toabctl:
+1

Itxaka:
+1000

nicolasbock:
+1

PR #2268, opened 2016-09-09
Cleanup the admin upgrade script

jsuchome:
Based on the changes done to the node upgrade script in https://github.com/crowbar/crowbar-core/pull/640

### DIFF
--- a/scripts/upgrade_admin_server.sh
+++ b/scripts/upgrade_admin_server.sh
@@ -18,17 +18,38 @@ exec 1>&4 2>&1
 
 set -x
 
+INSTALLDIR=/var/lib/crowbar/install
+RUNDIR=/var/run/crowbar
+RUNFILE=$RUNDIR/admin-server-upgrading
+
+cleanup()
+{
+    echo "cleaning up after interrupt or exit"
+
+    # cleanup upgrading indication so that the action can be restarted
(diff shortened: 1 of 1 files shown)


rsalevsky:
Maybe it's a stupid question but can we use the same template here too and create the file with chef?

jsuchome:
Well, the template can't be the same , although it's true it is similar.
We might use some template for admin server, but we'd need to run chef client to install it. Isn't packaging the script simpler?

rsalevsky:
It's is, it was more a question of what is easier to maintain? If we can use the same template we just need to update it in one place if something is broken otherwise we have always two different places. If we have it in two different places the chance is also high that we forget to update the other one.

jsuchome:
So are you talking about using same template we have for nodes also for admin server? I would probably not go there. The need for various checks in such templates, if it is node or admin specific would make it IMHO less readable.

Or should we use 2 templates, but at least we'd get rid of the script here. However we'd need to add it to some special role for admin server and run chef client for that case on the admin server. 
We might do it, but I'd vote for doing it in some cleanup/refactoring phase.

rsalevsky:
I'm fine with this change if you say it makes no sense and it would make it much more complex.

+1

vuntz:
Assuming it's tested, +1.

PR #2271, opened 2016-09-27
Set no_proxy in install-chef-suse.sh (bsc#998968)

s-t-e-v-e-n-k:
Chef now works around a missing Ruby feature by checking no_proxy
itself and matching against it when returning if a proxy connection
is required, so drop the http_proxy workaround and set no_proxy
directly. Since the script talks to both the hostname and the IP, add
them both.

### DIFF
--- a/scripts/install-chef-suse.sh
+++ b/scripts/install-chef-suse.sh
@@ -355,6 +355,9 @@ reset_crowbar
 FQDN=$(hostname -f 2>/dev/null);
 DOMAIN=$(hostname -d 2>/dev/null);
 IPv4_addr=$( getent ahosts $FQDN 2>/dev/null | awk '{ if ($1 !~ /:/) { print $1; exit } }' )
+# Set no_proxy for localhost, the FQDN and the hostname, so chef will not use
+# the proxy for them.
+export no_proxy="$no_proxy,localhost,$FQDN,$IPv4_addr"
 
 # Sanity checks
 # -------------
@@ -681,9 +684,6 @@ echo_summary "Performing initial chef-client run"
 service chef-client status &> /dev/null && service chef-client stop
 
 if ! [ -e ~/.chef/knife.rb -a -e ~/.chef/root.pem ]; then
-    # no_proxy is currently not supported in ruby see bsc#958716
-    # unset it for this call
-    (unset http_proxy
     if knife client list | grep -q "^ *root$"; then
         knife client delete --yes root
     fi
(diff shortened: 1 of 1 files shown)


rsalevsky:
AFAIK, there was no change in our chef codebase so I'm not sure why this should work now. Can you explain why you think it works now and what changed?

s-t-e-v-e-n-k:
Firstly, I checked the chef REST client code in Cloud 6, which has exclude functionality based on no_proxy, and secondly, I pulled up a development environment to the point where crowbar was installed, exported http_proxy to a machine on my local netwok on port 3128, ensuring any request would get connection refused, and made certain that install-suse-cloud worked.

rsalevsky:
Mmh... I tried exactly the same while I created the workaround but it didn't work for me. So I don't know if I did something wrong while testing or if something changed in the code.

s-t-e-v-e-n-k:
Testing this on mkcloud to be sure:

mkcloud:~/manual.stevek # ./runtest.c18.stevek cleanup prepare setupadmin prepareinstcrowbar
mkcloud:~/manual.stevek # ssh crowbar.c18.cloud.suse.de 
<apply patch>
root@crowbar:~ # nc -vz 10.162.191.4 3128
nc: connect to 10.162.191.4 port 3128 (tcp) failed: Connection refused
root@crowbar:~ # export http_proxy=http://10.162.191.4:3128/
root@crowbar:~ # echo $http_proxy 
http://10.162.191.4:3128/
root@crowbar:~ # echo $no_proxy

root@crowbar:~ # install-suse-cloud
<dialogs about starting services, running chef-client and installing barclamps>

This actually fails with:

Retrieving package bridge-utils-1.5-17.56.x86_64 (1/1),  32.1 KiB ( 66.0 KiB unpacked)
Retrieving: bridge-utils-1.5-17.56.x86_64.rpm [error]
Abort, retry, ignore? [a/r/i/? shows all options] (a): a
STDERR: Download (curl) error for 'http://149.44.176.43/repos/SLES12-SP1-Pool/x86_64/bridge-utils-1.5-17.56.x86_64.rpm':
Error code: Connection failed
Error message: Failed to connect to 10.162.191.4 port 3128: Connection refused

However, I think that's fine since in the usual case, a configured proxy server will be able to reach the SLES repos to download an rpm file, and I think it shows the proxy server isn't being used for the myriad of knife calls and then running chef-client.

rsalevsky:
Change looks good, but can you rebase to retrigger the gating?

s-t-e-v-e-n-k:
No other changes have merged, so a rebase is a no-op. However, I will amend HEAD to change the date and force push to do what you ask.

PR #2273, opened 2016-10-17
Set no_proxy in install-chef-suse.sh (bsc#998968)

s-t-e-v-e-n-k:
Chef now works around a missing Ruby feature by checking no_proxy
itself and matching against it when returning if a proxy connection
is required, so drop the http_proxy workaround and set no_proxy
directly. Since the script talks to both the hostname and the IP, add
them both.

Cherry-picked from 43d7b604bd7e8753dfbd5b31fb838b0087ea4adf.


vuntz:
Gating failed in CCT, so unrelated to the patch.

PR #2276, opened 2016-10-26
Delete OS values after the package upgrade

jsuchome:
This is an attempt to fix https://bugzilla.suse.com/show_bug.cgi?id=1006770

Needs backport to Cloud6

### DIFF
--- a/scripts/upgrade_admin_server.sh
+++ b/scripts/upgrade_admin_server.sh
@@ -77,6 +77,15 @@ upgrade_admin_server()
     # Signalize that the upgrade correctly ended
     echo "12.2" >> $UPGRADEDIR/admin-server-upgraded-ok
 
+    # delete current OS values
+    knife exec -E "n = nodes.find(:roles => 'provisioner-server').first
+n.target_platform = nil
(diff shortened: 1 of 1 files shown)


scripts/upgrade_admin_server.sh
```diff
@@ -77,6 +77,15 @@ upgrade_admin_server()
...
 
+    # delete current OS values
+    knife exec -E "n = nodes.find(:roles => 'provisioner-server').first
+n.target_platform = nil
+n.provisioner.default_os = nil
+n.save"
+
+    # let the provisioner fill the new OS values
+    chef-client
+
```
rhafer: Is the chef-client run here really required. Or could we just explictily set the attributes to the new values? This is somewhat along the lines of @MaximilianMeister comment (https://github.com/crowbar/crowbar/pull/2276#pullrequestreview-6998801)

jsuchome: Are you suggesting to use `suse-12.2` values directly or just asking what is better?

Chef-client run does set the values correctly and does not seem to do any harm.

PR #2277, opened 2016-10-27
travis: specify newer ruby version to use

MaximilianMeister:
otherwise 1.9.3 is used and gems can't be installed

see https://github.com/crowbar/crowbar/pull/2276

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,9 @@
-language: python
+language: ruby
 sudo: false
+
+rvm:
+  - 2.1.0
+
 install:
     - bundle install
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ rvm:
   - 2.1.0
 
 install:
+    - pip install bashate --user
     - bundle install
 script:
     - rake test:bashate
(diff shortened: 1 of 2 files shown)


MaximilianMeister:
still the old config is used for the job, so travis can't really be green here

jsuchome:
I would like to approve it, but I actually do not know if the value is correct, at the right place etc.

Itxaka:
@jsuchome https://docs.travis-ci.com/user/languages/ruby/#Choosing-Ruby-versions-and-implementations-to-test-against

Looks fine, but it didnt work :/
```
Pre-installed Ruby versions
ruby-1.9.3-p551
```

Itxaka:
Why is the language set to python? Due to that, travis is ignoring the rvm keyword as...why shouldn't it if it expects python :)


from https://docs.travis-ci.com/user/languages/ruby/#Choosing-Ruby-versions-and-implementations-to-test-against
```
Note that the rvm: key is only available in Ruby Build Environments, not in other images containing a ruby implementation.
```

MaximilianMeister:
@Itxaka wasn't that trivial to just switch the language. i guess the language was python because of bashate. weird setup... anyways fixed now. thanks for spotting this

PR #2279, opened 2016-11-02
add some hints on debugging a crowbar devsetup

MaximilianMeister:


doc/development.md
```diff
@@ -132,3 +132,29 @@ Before we can start you need to match some prerequirements on your host machine.
...
   7. Now you can access you crowbar development setup via ```http://your.crowbar.instance:5000```
+
+## Debugging
+
+If you want to debug some specific piece of code you can use `byebug` which is also mentioned in the `Gemfile`
+
+To set it up you have to install the package:
+
+```bash
+zypper ar -f http://dist.suse.de/install/SLP/SLE-12-SP2-SDK-LATEST/x86_64/DVD1/ sle12-sp2-sdk
```
rsalevsky: In the devsetup the byebug gem is installed via bundler and the SDK repo is added as well.

aspiers:
From now on I'm adding the #toomanyreviewers tag to PRs, to collect proof that we don't spread our reviewing efforts evenly. 5 in this case.

aspiers:
No need to wait for mkcloud gating here.

PR #2281, opened 2016-11-04
Update the OS values after the package upgrade of admin server

jsuchome:
An alternative to https://github.com/crowbar/crowbar/pull/2276. Backport for Cloud6 is needed


jsuchome:
[Bug 1006770 - during upgrade, default_os and target_platform set incorrectly](https://trello.com/c/TNUan7h4/167-bug-1006770-during-upgrade-default-os-and-target-platform-set-incorrectly)

PR #2286, opened 2016-11-24
[3.0] upgrade: make a backup of the important database state

MaximilianMeister:
in some occasions the files have disappeared for unknown reasons after
the admin server upgrade.
not only because of that it makes sense to copy them away to a safe
place.

### DIFF
--- a/scripts/upgrade_admin_server.sh
+++ b/scripts/upgrade_admin_server.sh
@@ -50,7 +50,9 @@ upgrade_admin_server()
     # we will need the dump for later migrating it into postgresql
     pushd /opt/dell/crowbar_framework
     sudo -u crowbar RAILS_ENV=production bin/rake db:migrate
+    cp -a db/schema.rb /var/lib/crowbar/upgrade
     sudo -u crowbar RAILS_ENV=production bin/rake db:dump
+    cp -a db/data.yml /var/lib/crowbar/upgrade
     popd
 
     ### Chef-client could lockj zypper and break upgrade


MaximilianMeister:
needs forward port

MaximilianMeister:
I can open another PR for importing those files from crowbar-init, but lets get this discussed/merged before

scripts/upgrade_admin_server.sh
```diff
@@ -50,7 +50,9 @@ upgrade_admin_server()
     # we will need the dump for later migrating it into postgresql
     pushd /opt/dell/crowbar_framework
     sudo -u crowbar RAILS_ENV=production bin/rake db:migrate
+    sudo -u crowbar cp db/schema.rb /var/lib/crowbar/upgrade
```
jsuchome: Why `sudo -u crowbar cp`? Coudln't we just `cp -a`?

jsuchome:
So... maybe it makes sense to make to backup anyway, but this seems to hide the real error...

scripts/upgrade_admin_server.sh
MaximilianMeister: ok. would be even better.

MaximilianMeister:
> So... maybe it makes sense to make to backup anyway, but this seems to hide the real error...

yes, backing up this file is really crucial here. we can add a big warning in crowbar-init when it can't find the file in the db folder so the real error won't be that hidden

jsuchome:
@MaximilianMeister could you please reference the bug in the commit message (https://bugzilla.suse.com/show_bug.cgi?id=1012177) and prepare the change for crowbar-init?

MaximilianMeister:
> @MaximilianMeister could you please reference the bug in the commit message (https://bugzilla.suse.com/show_bug.cgi?id=1012177) and prepare the change for crowbar-init?

@jsuchome done

PR #2287, opened 2016-11-29
upgrade: make a backup of the important database state (bsc#1012177)

MaximilianMeister:
in some occasions the files have disappeared for unknown reasons after
the admin server upgrade.
not only because of that it makes sense to copy them away to a safe
place.

(cherry picked from commit 884944c900f5de1d3ce089bdea6c859fbcc2854e)

### DIFF
--- a/scripts/upgrade_admin_server.sh
+++ b/scripts/upgrade_admin_server.sh
@@ -56,7 +56,9 @@ upgrade_admin_server()
     # we will need the dump for later migrating it into postgresql
     pushd /opt/dell/crowbar_framework
     sudo -u crowbar RAILS_ENV=production bin/rake db:migrate
+    cp -a db/schema.rb /var/lib/crowbar/upgrade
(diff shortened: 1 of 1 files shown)


jsuchome:
Not sure if it is really needed (we should have solved the real bug by the time of 7-8 upgrade). But it does not hurt, sure.

Itxaka:
as this may be resolved in the future, could you add a TODO or a FIXME to that line so we dont forget to remove it when we fix the real issue?

MaximilianMeister:
> Not sure if it is really needed (we should have solved the real bug by the time of 7-8 upgrade). But it does not hurt, sure.

@jsuchome i think we have to do it anyways, as it is important data which would be catastrophic if it got lost (and if there is no other backup)

> as this may be resolved in the future, could you add a TODO or a FIXME to that line so we dont forget to remove it when we fix the real issue?

@Itxaka I'd really keep it as a feature, as it's not really a full workaround

PR #2288, opened 2016-11-30
Add script to reindex the solr database

rsalevsky:
This script is required in case the solr database is stored in tmpfs.


aspiers:
Why is this required, and what calls it?

PR #2289, opened 2016-11-30
[3.0] Add script to reindex the solr database

rsalevsky:
This script is required in case the solr database is stored in tmpfs.

(cherry picked from commit 6fc690c3e57c49112c1492ced8421da9b5e768a7)

Backport of #2288


rsalevsky:
The script is not installed and the gating failed in tempest so it's unrelated.

PR #2290, opened 2016-12-03
Fix call to crowbar:schema_migrate rake task in barclamp_install.rb

vuntz:
Right now, the migrations are never run for most barclamps due to a bug:
the list of barclamps must be space-separated (with commas, rake would
filter items after the first one).


MaximilianMeister:
do we need a backport here?

vuntz:
> do we need a backport here?

No, in stable/3.0, we still do the schema migration one by one.

PR #2291, opened 2016-12-06
install-suse-cloud: log command line args

nicolasbock:
We need to move the logging of command line args _before_ we parse and
shift them.

### DIFF
--- a/scripts/install-chef-suse.sh
+++ b/scripts/install-chef-suse.sh
@@ -47,6 +47,9 @@ EOF
     exit
 }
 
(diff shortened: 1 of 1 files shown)


nicolasbock:
@MaximilianMeister that's a good point. I redid the patch a bit. What do you think?

nicolasbock:
@rsalevsky Sorry, I keep forgetting about that format :)

nicolasbock:
Backport at https://github.com/crowbar/crowbar/pull/2293

PR #2294, opened 2016-12-20
Barclamp guide

jgrassler:
This pull requests adds a step-to-step guide for writing a new barclamp.

### DIFF
--- a/doc/barclamp.md
+++ b/doc/barclamp.md
@@ -1,0 +1,1358 @@
+# Barclamp development guide
+
+This document is an introduction to Barclamp development. It will guide you
+through all the steps neccessary to create a minimal (without UI elements)
+Barclamp for Crowbar. It will then guide you through all the testing, CI and
+review steps until the point where it gets merged into Crowbar.
+
+## What is a Barclamp?
+
+A barclamp is a plugin for Crowbar that configures a particular service or
+group of services. It contains a Chef cookbook with one or more roles to
+perform the actual configuration work and Crowbar UI code to parametrize this
+chef cookbook and orchestrate applying its roles to individual machines.
+
+## Terms and definitions
+
+We will assume throughout this text that your barclamp is will be named
+`mybarclamp`. Please substitute your actual barclamp's name for this
+placeholder when you create a barclamp. The convention for OpenStack barclamps
+is to use the OpenStack component's name, e.g. `nova`.
+
+## Development Environment
+
+This guide describes a very basic remote development environment on a Crowbar
+admin node. You may skip this section entirely if you prefer working locally
+(in this case you might want to look at our
+[Guard based approach](https://github.com/crowbar/crowbar/blob/master/Guardfile)
+and/or
+[build-dist-rpms](https://github.com/openSUSE/pack-tools/tree/master/contrib/BS-pkg-testing)
+for local development), or already use an entirely different development
+environment.
+
+The environment in this section serves three chief purposes:
+
+* Provide something anybody can use even if their workstation accidently got an
+  anvil dropped on it an hour ago and now they have a new machine fresh from IT
+  with none of their customized personal environment on it. A `ssh` client is
+  all you need for this guide.
+
+* Document all the steps you need to go through to move your barclamp from a
+  git working copy a point where it runs in Crowbar. This is helpful for both
+  using (and debugging) existing development tools and for
+
+* Last but not least it provides us with the handy `reposync_crowbar` command
+  we will be using throughout this document. If you are using a different
+  development environment, please substitute "ensure your code's current state
+  of development is active in Crowbar" wherever this command occurs in the
+  text.
+
+With the example development environment, most of your development will take
+place on your Crowbar node (log in as root).  You will find the currently
+running Crowbar code (both Rails application components and chef recipes) in
+the `/opt/dell` directory. We will provide detailed locations for all important
+components of a Barclamp in the _Basic Implementation_ section below.
+
+In order to track changes you make we recommend you keep a checkout of your
+`crowbar-openstack` fork on the Crowbar node and apply the changes you make in
+there
+
+
+### Prerequisites
+
+We will assume some things to already be present. Namely, the following:
+
+* A Github fork of the appropriate Barclamp collection. For instance, you will
+  find OpenStack related barclamps in the
+  [crowbar-openstack](https://www.github.com/crowbar/crowbar-openstack)
+  repository and high availability related Barclamps in the
+  [crowbar-ha](https://github.com/crowbar/crowbar-ha) repository. For the sake
+  of readability we will assume you are developing an OpenStack related and
+  forked [crowbar-openstack](https://www.github.com/crowbar/crowbar-openstack)
+  throughout the rest of this document.
+
+* A Crowbar admin node with the Crowbar version you are developing against
+  (for new barclamps this will be `master`) already installed and running. This
+  should be as up-to-date as possible to reduce the chance of merge conflicts
+  happening.
+
+* A set of machines for your Crowbar admin node to orchestrate. In the case of
+  SUSE OpenStack Cloud this would be a controller and one compute node at a
+  minimum.
+
+* You will need to install the following packages:
+
+  * `patch`
+  * `git-core`
+
+Providing these is out of scope for this document. We assume you are able to
+fork the appropriate Barclamp collection. The machines can be provided through
+manual installation or by using an automated setup tool such as
+[mkcloud](https://github.com/SUSE-Cloud/automation/blob/master/scripts/mkcloud).
+Depending on the individual situation you can use either the `plain` or the
+`instonly` `mkcloud` step. The former will install the nodes and deploy the
+standard set of barclamps in their default configuration, while the latter will
+only install the nodes. Use `plain` if your barclamp does not need to be
+deployed before other barclamps and if you do not modify other barclamps.
+Otherwise use `instonly` and deploy barclamps bit by bit through the Crowbar
+web UI or the Crowbar CLI client. This will give you the opportunity to add
+package repositories or tweak barclamps' proposal settings where needed along
+the way.
+
+Optionally (but strongly recommended) you can also
+
+* Create a backup of your Crowbar admin node: You are probably going to break
+  it a few times in the course of testing your new barclamp so it is a good
+  idea to have a backup you can quickly restore rather than rebuild your entire
+  test environment. If you are using `mkcloud` you can create and restore a LVM
+  snapshot of your admin node using its `createadminsnapshot` and
+  `restoreadminfromsnapshot` steps, respectively.
+
+* Create an additional OpenStack controller or compute node as a scratch node:
+  this will allow you to deploy only your own barclamp's controller side
+  components to this node while all other OpenStack services reside on the main
+  controller. This greatly shortens the time a chef run for your barclamp's
+  controller side components takes and thus speeds up development. Make a
+  backup of this node as well to guard against breakage and allow for
+  clean-slate deployment of your barclamp's controller side components.
+
+The best way to set up your scratch node is to make it a lonely node. To this
+end, you will need to re-run mkcloud after you have created your cloud:
+
+```
+nodenumberlonelynode=1 mkcloud setuplonelynodes crowbar_register
+
+```
+
+### Preparations for Crowbar Node
+
+These are recommendations, but we'll assume you followed them for the remainder
+of this document since this shortens the commands in our examples. First of
+all, clone your `crowbar-openstack` fork to `/root/crowbar-openstack` on your
+Crowbar node and switch to your development branch. Note, that the code below
+assumes you are working on a dedicated topic branch. It will break if you are
+doing development on your fork's `master` branch. So do not work on `master`!
+
+Next, create the following /root/.bashrc with a few useful aliases and
+environment variables (substitute with appropriate values where so indicated by
+comments):
+
+```
+# Replace "Crowbar Committer" by your name as it should appear in commit
+# messages.
+C_NAME="Crowbar Committer"
+
+# Replace "email@address.invalid" by your email address as it should appear in
+# commit messages.
+C_EMAIL="email@address.invalid"
+
+# Insert your own barclamp's name here:
+export BARCLAMP_NAME=mybarclamp
+
+# Insert a space delimited list of the Barclamps you are working on here (only
+# required if your barclamp requires changes to other barclamps as well).
+
+export BARCLAMPS="$BARCLAMP_NAME"
+
+# Insert the path to your barclamp collection checkout here (in case it's
+# somewhere other than /root/crowbar-openstack). This directory must be a git
+# working copy of the crowbar-openstack fork and branch you are working on.
+
+export BARCLAMP_FORK_PATH=/root/crowbar-openstack
+
+export GIT_AUTHOR_NAME="$C_NAME"
+export GIT_AUTHOR_EMAIL="$C_EMAIL"
+export GIT_COMMITTER_NAME="$C_NAME"
+export GIT_COMMITTER_EMAIL="$C_EMAIL"
+
+# Applies all the committed changes in your crowbar-openstack working copy to
+# /opt/dell and updates Crowbar runtime state. Note: this will break if you
+# rebase your working copy, so do not rebase your working copy. If you rebase
+# it (e.g. for final testing before a merge, please build a fresh Crowbar node
+# or restore it from a snapshot).
+function reposync_crowbar()
+  {
+  if [ -e /opt/dell/patchlevel ]; then
+    . /opt/dell/patchlevel
+    head=$(cd $BARCLAMP_FORK_PATH; git log --format='%H'| head -n 1)
+    if [ "$head" = "$commit" ]; then
+      echo "/opt/dell is already up to date. If this is not what you " 1>&2
+      echo "expected to see, you may have uncommitted changes" 1>&2
+      echo "in ${BARCLAMP_FORK_PATH}." 1>&2
+      return
+    fi
+  fi
+
+  pushd $BARCLAMP_FORK_PATH > /dev/null
+
+  # This backup preserves the original /opt/dell from the RPM
+  if [ ! -d /opt/dell.orig ]; then
+    cp -a /opt/dell /opt/dell.orig
+  fi
+
+  # This backup preserves the /opt/dell from before the current
+  # reposync_crowbar run. It is only preserved if reposync_crowbar
+  # fails.
+  if [ ! -d /opt/dell.bak ]; then
+    cp -a /opt/dell /opt/dell.bak
+  else
+    echo "Found a previous intermediate backup of /opt/dell in /opt/dell.bak" 1>&2
+    echo "This indicates a failed reposync_crowbar run." 1>&2
+    echo "Please restore /opt/dell to a known good state from either of" 1>&2
+    echo "    /opt/dell.bak" 1>&2
+    echo " or /opt/dell.orig" 1>&2
+    echo " and remove /opt/dell/back.bak once you have done so." 1>&2
+    return 1
+  fi
+  # Note: $commit will only be set when sourced from /opt/dell/patchlevel.
+  if [ -n "$commit" ]; then  # only apply commits added since the last reposync_crowbar run
+    git diff $commit HEAD | patch -N --merge -p1 -d /opt/dell || return 1
+  else # apply all commits in topic branch
+    git diff origin/master...HEAD | patch -N --merge -p1 -d /opt/dell || return 1
+  fi
+
+  # Record current HEAD as patchlevel
+  echo "commit=$(git log --format='%H'| head -n 1)" > /opt/dell/patchlevel
+  popd > /dev/null
+
+  for barclamp in $BARCLAMPS
+    do
+    if [ ! -h /opt/dell/crowbar_framework/barclamps/${barclamp}.yml ]; then
+      ln -s /opt/dell/${barclamp}.yml /opt/dell/crowbar_framework/barclamps 2>&1 > /dev/null
+    fi
+    done
+  chmod 755 /opt/dell/bin/*
+
+  sync_crowbar
+  rm -rf /opt/dell.bak
+  }
+
+# Helper function: updates Crowbar's runtime state after /opt/dell
+# has been modified.
+sync_crowbar()
+  {
+  barclamp_install.rb /opt/dell/crowbar_framework/barclamps &&
+  for barclamp in $BARCLAMPS
+    do
+    knife cookbook upload -o /opt/dell/chef/cookbooks/ $barclamp
+    done
+  systemctl restart crowbar
+  }
+
+```
+
+We will make heavy use of the `reposync_crowbar` command thus defined in the
+following sections. A short explanation of the three:
+
+`reposync_crowbar` will apply your whole fork against `/opt/dell` using
+`patch(1)` and sync Crowbar's runtime state with what is now in /opt/dell. You
+use this command to apply the current state of your fork to a freshly installed
+Crowbar node. This command may cause merge conflicts which you will have to
+resolve before proceeding. They usually happen due to a very out of date topic
+branch. While you can resolve them in /opt/dell you are better off rebasing
+your `crowbar-openstack` fork against upstream and applying the result to a
+fresh Crowbar node. Also, you will find a backup copy of /opt/dell in
+/opt/dell.orig if a merge conflict does occur.
+
+Using patches against `/opt/dell` may seem a bit over-the-top at first since
+one could simply copy files back and forth or use `barclamp_install` on your
+checkout, but it prevents a lot of breakage in cases where `/opt/dell` is out
+of sync with the `crowbar-openstack` repository's master branch or your own
+fork of `crowbar-openstack` is out of sync with upstream.
+
+In the former case copying files from your fork to `/opt/dell` or using
+`barclamp_install` may cause an inconsistent state and errors entirely
+unrelated to your change.
+
+In the latter case copying files from `/opt/dell` to your fork may re-introduce
+old code that is no longer present upstream. Both should not be an issue for
+new barclamps, but it is better to get used to the `patch(1)` based workflow
+right away since these problems do become a problem when changing an existing
+barclamp later and thus the `patch(1)` based workflow becomes neccessary.
+
+## Packages
+
+If you create a new barclamp this is typically to deploy an OpenStack service
+that does not exist on SUSE OpenStack Cloud, yet. While important, your
+barclamp is only half of what is needed. The other half is creating the
+packages your Barclamp will install the new OpenStack service from.
+
+Ideally, you have already created these packages in [Open Build
+Service](https://build.opensuse.org) and have taken the appropriate steps to
+get them into the repository `mkcloud` uses for deployment (since `mkcloud`
+uses SUSE internal repositories or a local copy of these internal repositories,
+there is no point in detailing these steps in this public guide). In this case
+you are all set, and your chef recipes will merrily install the packages they
+need from official sources. If this isn't the case for you there are still some
+strategies for getting your packages installed:
+
+First of all, you will need a package repository to store your non-official
+packages in. The easiest way to accomplish this is to register an account on
+[Open Build Service](https://build.opensuse.org) and create a project with all
+the packages you need under your home project. We'll assume you have this a
+non-official package repository (or multiple repositories) with all the
+packages you need in place for the remainder of this section.
+
+If all you do is adding new packages, you should be fine with simply adding
+your repository using `zypper ar` on every node you are installing your
+packages to:
+
+```
+zypper ar -f http://download.opensuse.org/repositories/home:/<your OBS user>:/<your project name>/SLE_12_SP2 myrepo
+zypper refresh
+```
+
+If you need to override existing packages with newer versions (for instance, if
+the service your barclamp deploys requires a newer python-openstackclient
+version you ship in your own repository), you may also need to disable vendor
+locking for Zypper. To do this, set the following in `/etc/zypper/zypp.conf`:
+
+```
+solver.allowVendorChange = true
+
+```
+
+You need to do this because normally Zypper would prefer packages from official
+repositories, even if your own repository contains newer version. If you set
+this option the newer package from your repository will trump the one from
+the official one.
+
+Both steps must be performed on all relevant nodes _before_ you apply your
+barclamp, otherwise applying it will fail due to Crowbar being unable to locate
+your packages.
+
+## Barclamp Component Overview
+
+A barclamp consists of two main components:
+
+* A [chef cookbook](https://docs.chef.io/cookbooks.html) which takes care of
+  actually deploying your service.
+* Some glue code to tie your chef cookbook into the Crowbar Web UI and backend.
+  This glue code will parametrize your chef cookbook and apply its component
+  roles to the nodes it orchestrates.
+
+The subsequent sections will describe these two components and the ingredients
+that go into them into detail. They start of with an overview of a barclamp's
+development phases as we know them from experience and take you through all
+aspects of barclamp implementation in the order you are likely to encounter
+them. They are meant to be read in parallel to developing your own barclamp.
+
+## Development Phases
+
+In this section we will show a development workflow adapted for the remote
+development environment we described in the beginning. This is not the
+end-all-be-all of Barclamp development workflows, but it should be useful for
+new developers. If you do not like this approach (for instance, if you prefer
+developing in your local environment), our [Guard based
+approach](https://github.com/crowbar/crowbar/blob/master/Guardfile) provides an
+alternative. If your patch nears readiness, you may also find
+[build-dist-rpms](https://github.com/openSUSE/pack-tools/tree/master/contrib/BS-pkg-testing)
+useful. This tool will let you quickly build RPMs from your `crowbar-openstack`
+fork.
+
+The Development workflow essentially consists of 3 phases: 
+
+1. Basic Implementation: In this phase you will write a scaffold for the
+   Barclamp's chef code, define roles and parameters and create UI code for
+   parametrizing your Barclamp. Note that no testing needs to take place during
+   this phase (and doesn't make much sense either). This phase produces a rough
+   draft that may have any number of errors but will already define your
+   barclamp's basic architecture.
+
+2. Local Testing: in this phase you will debug and fix the rough draft from the
+   first phase until the UI side of the barclamp works well and parametrizes
+   the barclamp as expected. As a second step you will fill your chef recipe
+   scaffold with life and develop your chef recipes to the point where they
+   correctly deploy what they are supposed to deploy.
+
+3. Pull Request Testing: once you are finished with phase 2 to the point where
+   the barclamp performs to your own satisfaction you submit a pull request for
+   your barclamp against the relevant Crowbar repository's `master` branch
+   (e.g. against the `crowbar-openstack` `master` branch). Now Crowbar's CI
+   tests will run using the code from your pull request and Crowbar maintainers
+   will review it. During this phase you may again have to fix a number of
+   things and may well need to return to phase 2 for local testing of the
+   changes you make (it is not recommended to rely on CI tests by themselves
+   for extensive changes).
+
+Below we will describe these phases in detail.
+
+_Note: unless otherwise indicated, all paths from here on out are relative to
+your_ `crowbar-openstack` _fork's root directory._
+
+### Basic implementation
+
+In this phase you will do a scaffold implementation of your barclamp. This
+means that you will mostly focus on the Crowbar UI side of things and how it
+ties into your barclamp's chef cookbook. This means that on the chef side
+you'll only implement the structure of your chef cookbook, i.e. define the
+parameters it takes and the roles it defines for the Crowbar UI to use.
+
+You should also create recipes and configuration templates in your cookbook
+already, but you only need to flesh these out as far as you need them to be
+clear about the parameters you'll need Crowbar to supply.
+
+Throughout this phase you can work on your local machine or on your development
+Crowbar admin node, as you prefer, since the testing you can do in this phase is
+limited to basic ruby syntax checks using `ruby -c`. Once you are finished with
+this phase development will have to move to your Crowbar admin node, though.
+
+#### Chef Cookbook
+
+Location: `chef/cookbooks/`
+
+The best point to start barclamp development is the
+chef cookbook. Each barclamp contains
+one of these. The chef cookbook is your barclamp's core component. It contains
+the code that will do the grunt work of deploying the services, configuration
+files and runtime state handled by your barclamp. As a first step, create
+`chef/cookbooks/` now.
+
+#### Cookbook metadata
+
+Locations:
+
+* `chef/cookbooks/mybarclamp/metadata.rb`
+* `chef/cookbooks/mybarclamp/README.md`
+
+Before we begin with the cookbook proper we will need to take care of some
+boilerplate files. The first of these is `metadata.rb`. It mostly contains
+various descriptive information about your barclamps. Here's an example from
+the Barbican barclamp (`chef/cookbooks/barbican/metadata.rb`):
+
+```
+name "barbican"
+maintainer "Crowbar project"
+maintainer_email "crowbar@googlegroups.com"
+license "Apache 2.0"
+description "Installs/Configures Barbican"
+long_description IO.read(File.join(File.dirname(__FILE__), "README.md"))
+version "0.1"
+
+depends "apache2"
+depends "database"
+depends "keystone"
+depends "crowbar-openstack"
+depends "crowbar-pacemaker"
+depends "utils"
+```
+
+The contents of these fields are mostly free-form, hence you can copy the file
+from another barclamp and simply adapt the `name` and `description` fields. The
+one exception to that rule are the `depends` statements: these define which
+other barclamps your own barclamp depends on because it uses their cookbooks'
+contents. If, for instance, you use the Keystone barclamp's
+`KeystoneHelper.keystone_settings()` method (which you should for OpenStack
+barclamps), you'd add a `depends "keystone"` statement here.
+
+You may already have noticed the second piece of boilerplate: the README.md
+file referenced by the `long_description` statement. This goes one goes into
+`chef/cookbooks/` and describes your Chef cookbook as
+verbosely or tersely as you like (please put at least one complete sentence in
+there).
+
+#### Recipes
+
+Location: `chef/cookbooks/mybarclamp/recipes/*.rb`
+
+[Recipes](https://docs.chef.io/recipes.html) make up the bulk of your chef
+cookbook's payload. These recipes are written in Ruby (hence they must have a
+`.rb` extension) and are the smallest organization unit of a chef cookbook.
+They consist of [resource](https://docs.chef.io/resource.html) definitions that
+define various aspects of system state, such as a configuration file's content,
+packages to install, or a service to enable and start. In this step the recipes
+do not need to contain any code, yet. You primarily need to know which recipes
+to define and what each recipe's purpose is at this stage.
+
+You are free to organize your recipes as you see fit. There are no
+hard-and-fast rules for this since they mostly depend on the application your
+barclamp deploys:
+
+For instance, if your application (like most OpenStack components) consists of
+an API service and a RabbitMQ driven backend service, you might want to split
+these into a `api.rb` and `backend.rb` service. Likewise, if you need an
+agent to run on compute nodes, this might warrant a dedicated `agent.rb`
+recipe. If you want to use Crowbar's high availability features, it might also
+make sense to have a dedicated `ha.rb` recipe for users deploying it in a
+highly available manner.
+
+On a more general note, feel free to take a look at the other cookbooks'
+recipes in `chef/cookbooks/` for inspiration. We have barclamps for a range of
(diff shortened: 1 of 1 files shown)


doc/barclamp.md
```diff
@@ -220,21 +221,21 @@ this phase development will have to move to your Crowbar admin node, though.
...
-`chef/cookbooks/mybarclamp` now.
+`chef/cookbooks/` now.
 
 #### Cookbook metadata
 
 Locations:
 
-* `chef/cookbooks/mybarclamp/metadata.rb`
-* `chef/cookbooks/mybarclamp/README.md`
+* `chef/cookbooks//metadata.rb`
```
matrixik: Looks like here and in many other places you accidentally removed `mybarclamp` in this commit.

doc/barclamp.md
```diff
@@ -0,0 +1,1177 @@
+# Barclamp development guide
+
+This document is an introduction to Barclamp development. It will guide you
```
toabctl: A short description what a ```barclamp``` is would be nice in the intro

doc/barclamp.md
```diff
@@ -0,0 +1,1177 @@
...
+    knife cookbook upload -o /opt/dell/chef/cookbooks/ mybarclamp
+    done
+  systemctl restart crowbar
+  }
+```
+
+We will make heavy use of the `Cpatch`, `Fpatch` and `sync_crowbar` commands
+thus defined in the following sections. A short explanation of the three:
+
+* `Fpatch` will apply your whole fork against `/opt/dell` using `patch(1)`. You
```
toabctl: Why not using ```barclamp_install``` with a given directory?

doc/barclamp.md
```diff
@@ -0,0 +1,1177 @@
...
+  code will parametrize your chef cookbook and apply its component roles to the
+  nodes it orchestrates.
+
+The subsequent sections will describe these two components and the ingredients
+that go into them into detail. They start of with an overview of a barclamp's
+development phases as we know them from experience and take you through all
+aspects of barclamp implementation in the order you are likely to encounter
+them. They are meant to be read in parallel to developing your own barclamp.
+
+## Development Phases
```
toabctl: Do you know that there is a (broken) script to create a new barclamp? See https://github.com/crowbar/crowbar/blob/master/scripts/barclamp_create.rb .

doc/barclamp.md
```diff
@@ -0,0 +1,1177 @@
...
+# Replace "6cefbd80254141a63f4c86df195cd930ac7eb10b" by your first commit's ID.
+export COMMIT="6cefbd80254141a63f4c86df195cd930ac7eb10b"
+
+# Insert a space delimited list of the Barclamps you are working on here.
+export BARCLAMPS="nova magnum"
+
+export GIT_AUTHOR_NAME="$C_NAME"
+export GIT_AUTHOR_EMAIL="$C_EMAIL"
+export GIT_COMMITTER_NAME="$C_NAME"
+export GIT_COMMITTER_EMAIL="#C_EMAIL"
```
samirjorina: Should be "$C_EMAIL" instead of "#C_EMAIL"

doc/barclamp.md
```diff
@@ -0,0 +1,1177 @@
...
+
+alias Cpatch="(cd ~/crowbar-openstack; git show | patch -p1 --merge -d /opt/dell)"
```
samirjorina: I'd add -N to the patch command, so that it won't fail if patch was already applied

doc/barclamp.md
```diff
@@ -0,0 +1,1177 @@
...
+    done
+  chmod 755 /opt/dell/bin/*
+  }
+
+sync_crowbar()
+  {
+  barclamp_install.rb /opt/dell/crowbar_framework/barclamps/ &&
+  for barclamp in $BARCLAMPS
+    do
+    knife cookbook upload -o /opt/dell/chef/cookbooks/ mybarclamp
```
samirjorina: use $barclamp variable instead of hard-coded mybarclamp

matrixik:
I think it would also be valuable to have short section about where to place rpm for development.

doc/barclamp.md
```diff
@@ -0,0 +1,1177 @@
...
+these changes without the test automation pull request from the next section.
+
+Once they are happy they will approve your pull request. Your pull request will
+need two approvals by Crowbar team members to get merged. Before it gets
+merged, please rebase your pull request so it only consists of a single commit.
+
+#### Adding Test Automation
+
+To include your barclamp in `mkcloud` test automation you would fork the
+[automation]() repository, add your barclamp to the list of barclamps enabled
```
matrixik: Missing link.

doc/barclamp.md
```diff
@@ -0,0 +1,1177 @@
...
+Optionally (but strongly recommended) you can also
+
+* Create a backup of your Crowbar admin node: You are probably going to break
+  it a few times in the course of testing your new barclamp so it is a good
+  idea to have a backup you can quickly restore rather than rebuild your entire
+  test environment. If you are using `mkcloud` you can create and restore a LVM
+  snapshot of your admin node using its `createadminsnapshot` and
+  `restoreadminfromsnapshot` steps, respectively.
+
+* Create an additional OpenStack controller or compute node as a scratch node:
```
matrixik: Best way for creating this additional node?
Could lonely node be used for this?
If https://github.com/SUSE-Cloud/automation/pull/1549 will be accepted and merged then lonely node for development could be created with:
```
sudo nodenumberlonelynode=1 /path/to/mkcloud setuplonelynodes crowbar_register
```

doc/barclamp.md
```diff
@@ -0,0 +1,1177 @@
...
+  this will allow you to deploy only your own barclamp's controller side
+  components to this node while all other OpenStack services reside on the main
+  controller. This greatly shortens the time a chef run for your barclamp's
+  controller side components takes and thus speeds up development. Make a
+  backup of this node as well to guard against breakage and allow for
```
jgrassler: I'll add this as soon as there actually is generic backup-just-the-node-I-specified  and restore-just-the-node-I-specified functionality in `mkcloud`. I plan on adding this, but so far it does not exist, yet :-)

doc/barclamp.md
jgrassler: No, that one is news to me :-)

I'll have a look, thanks!

doc/barclamp.md
```diff
@@ -0,0 +1,1177 @@
...
+    - 'openstack'
+
+crowbar:
+  # The barclamp's priority in the crowbar UI (lower number means higher priority)
+  run_order: 98
+  # The barclamp's priority in the ordering of chef cookbooks (lower number
+  # means higher priority)
+  chef_order: 98
+
+  # FIXME: The following appear to be unused. They are neither loaded in
```
jgrassler: For now, yes :-)

More to the point, I want to leave it in there until either somebody who either knows more about Crowbar than me confirms these keys can indeed be omitted or I have tried it. Until then I prefer to leave the FIXME in place so I know this still needs addressing.

doc/barclamp.md
jgrassler: Done.

doc/barclamp.md
jgrassler: Fixed, thank you!

doc/barclamp.md
jgrassler: Good idea, thanks. Added.

doc/barclamp.md
jgrassler: Good point. Fixed. Also, I did the same for `/root/crowbar-openstack` while I was at it.

doc/barclamp.md
```diff
@@ -0,0 +1,1177 @@
...
+* `Fpatch` will apply your whole fork against `/opt/dell` using `patch(1)`. You
+  use this command to apply the current state of your fork to a freshly
+  installed Crowbar node. This command may cause merge conflicts. If this
+  happens, resolve them before doing anything else.
+
+* `Cpatch` will apply the latest commit in `/root/crowbar-openstack` to
+  `/opt/dell` (use this after each new commit). This command may cause merge
+  conflicts. If this happens, resolve them before doing anything else.
+
+* `sync-crowbar` will make any updated state in `/opt/dell` appear in the
```
jgrassler: Thanks, fixed!

doc/barclamp.md
```diff
@@ -0,0 +1,1177 @@
...
+* "Any OpenStack controller node"
+
+* "Any OpenStack compute node"
+
+* "Any Node in the Ceph Cluster"
+
+* "Any Ceph OSD node"
+
+We have a limited range of node types in Crowbar to define types of roles.
+Beyond that
```
jgrassler: Ah, thanks. I left that for "later" because I first needed to look up a bunch of URLs. Guess I forgot about it later...fixed.

doc/barclamp.md
```diff
@@ -0,0 +1,1177 @@
...
+* `service_user`
+
+When you are done `attributes[""]` should look something like this:
+
+```
+     "" : {
+        "api" : {
+           "bind_host" : "*",
+           "bind_port" : 12345,
+           "logfile" : "/var/log/myservice/myservice-api.log",
```
jgrassler: Done.

doc/barclamp.md
```diff
@@ -0,0 +1,1177 @@
...
+
+* `element_run_list_order`: this field governs the priority of your roles in
+  relation to the other roles assigned to any given node your own role is
+  assigned to. In other words, this governs local execution order of all roles
+  (i.e. not just yours) assigned to a given node. All of your roles should be
+  assigned a priority here.
+
+When you are done `deployment[""]` should look something like this
+(this example assumes you have defined a `mybarclamp-compute` and a
+`-controller` role):
```
jgrassler: Done.

doc/barclamp.md
```diff
@@ -0,0 +1,1177 @@
...
+to your barclamp's name, which will leave you with something like this for our
+example barclamp:
+
+```
+require File.join(File.expand_path(File.dirname(__FILE__)), "barclamp_lib")
+@barclamp = "mybarclamp"
+@timeout = 3600
+
+main
+```
```
jgrassler: Done.

doc/barclamp.md
```diff
@@ -0,0 +1,1177 @@
...
+* `crowbar_framework/config/locales/mybarclamp/en.yml`
+
+Since we are building a minimal barclamp without much in the way of UI
+components, we can stick to the bare minimum for the "View" part of MVC. We
+only need its components to be present without much content.
+
+First of all we create the directories
+
+* `crowbar_framework/app/views/barclamp/mybarclamp`
+* and `crowbar_framework/config/locales/mybarclamp`.
```
jgrassler: Agreed. That just sets people up for an `EPASTE`. Removed.

doc/barclamp.md
jgrassler: Looks good. Thanks :-)

I added this to the documentation in slightly modified form.

doc/barclamp.md
```diff
@@ -0,0 +1,1177 @@
...
+First of all, click on the `Create` button for your Barclamp in the list of
+barclamps. If this fails the reason is usually due to one of the following:
+
+* Errors in `crowbar_framework/app/models/mybarclamp_service.rb`
+* Syntax or naming errors in `crowbar_framework/app/controllers/mybarclamp_controller.rb`
+  (very uncommon since it's just boilerplate that can be produced through a
+  simple search-and-replace operation on another barclamp's controller)
+* `crowbar_framework/app/views/barclamp/mybarclamp/_edit_attributes.html.haml`
+* Errors in your default data bag, e.g.`chef/data_bags/crowbar/template-mybarclamp.json`
+  schema`. This is what error messages about invalid JSON syntax usually refer
```
jgrassler: Fixed.

doc/barclamp.md
jgrassler: Fixed.

doc/barclamp.md
jgrassler: Ah. That would do the trick, too, and be a lot simpler to boot. Are you currently using this approach? If so, are there any snags to watch out for?

doc/barclamp.md
```diff
@@ -5,6 +5,13 @@ through all the steps neccessary to create a minimal (without UI elements)
 Barclamp for Crowbar. It will then guide you through all the testing, CI and
 review steps until the point where it gets merged into Crowbar.
 
+## What is a Barclamp?
+
+A barclamp is a plugin for Crowbar that configures a particular service or
+group of services. It contains a che cookbook with one or more roles to
```
matrixik: a Chef cookbook

jgrassler: Thanks, fixed.

doc/barclamp.md
jgrassler: Hmm. Now that I think about it again, it's got the same problem as the copy-stuff-to-`/opt/dell` approach: if one doesn't rebase religiously and frequently, old code in _other_ barclamps that is no longer around in `master` might creep in, leading to all sorts of weird problems that have nothing to do with the barclamp being developed.

For completeness' sake I mentioned the `barclamp_install` approach alongside the `cp` one now.

aspiers:
Thanks for this!  When it is merged, please can you add a link from https://github.com/SUSE/cloud/tree/master/docs#crowbar ?

doc/barclamp.md
```diff
@@ -0,0 +1,1224 @@
...
+  for f in /opt/dell/*.yml
+    do
+    ln -s $f /opt/dell/crowbar_framework/barclamps 2>&1 > /dev/null
+    done
+  chmod 755 /opt/dell/bin/*
+  }
+
+sync_crowbar()
+  {
+  barclamp_install.rb $BARCLAMP_FORK_PATH &&
```
samirjorina: Here I'm using
`barclamp_install.rb /opt/dell/crowbar_framework/barclamps &&`

Otherwise I ended up with completely broken crowbar (problems with cinder, keystone, some weird migrations...). The suggested change simply 'reinstalls' existing barclamps and does not install them from git fork.

doc/barclamp.md
```diff
@@ -0,0 +1,1224 @@
...
+  for barclamp in $BARCLAMPS
+    do
+    knife cookbook upload -o /opt/dell/chef/cookbooks/ $BARCLAMP_NAME
```
samirjorina: I guess this should be:
`knife cookbook upload -o /opt/dell/chef/cookbooks/ $barclamp`

jgrassler: @aspiers: sure, will do!

@samirjorina: yes, there's a bit of a problem here indeed. Two problems actually:

The line you commented on causes the cookbook upload to fail completely. I don't think that would have caused the breakage you are observing, though.

That one is likely due to the second problem: I tried [Tom's suggestion](https://github.com/crowbar/crowbar/pull/2294#discussion_r93575431) a while ago and forgot to revert the `$BARCLAMP_FORK_PATH` in line 143  back to `/opt/dell/crowbar_framework/barclamps` before the next commit, effectively leaving the approach I tried there in place. Hence your problems with Cinder, Keystone and weird migrations are likely to be the situation described here: https://github.com/crowbar/crowbar/pull/2294#discussion_r94596298

Thank you for trying this and spotting the problem! I fixed both now.

doc/barclamp.md
```diff
@@ -0,0 +1,1224 @@
...
+# Replace "Crowbar Committer" by your name as it should appear in commit
+# messages.
+C_NAME="Crowbar Committer"
+
+# Replace "email@address.invalid" by your email address as it should appear in
+# commit messages.
+C_EMAIL="email@address.invalid"
+
+# Replace "6cefbd80254141a63f4c86df195cd930ac7eb10b" by your first commit's ID.
+export COMMIT="6cefbd80254141a63f4c86df195cd930ac7eb10b"
```
aplanas: What is this for?

jgrassler: Yes, it's used to create the diff. One could use something like `git log master..HEAD --oneline | tail -n 1 | cut -f 1`, I suppose, but that relies on the state of the `master` branch being somewhat sensible: if somebody did their changes in their fork's `master` branch or merged part of their topic branch's history to their fork's `master` branch it will blow up. People being people this is likely to happen at some point, and a hardwired commit ID will still Just Work if somebody messes up their local `master` branch.

doc/barclamp.md
```diff
@@ -0,0 +1,1224 @@
...
+* A Github fork of the appropriate Barclamp collection. For instance, you will
+  find OpenStack related barclamps in the
+  [crowbar-openstack](https://www.github.com/crowbar/crowbar-openstack)
+  repository and high availability related Barclamps in the
+  [crowbar-ha](https://github.com/crowbar/crowbar-ha) repository. For the sake
+  of readability we will assume you are developing an OpenStack related and
+  forked [crowbar-openstack](https://www.github.com/crowbar/crowbar-openstack)
+  throughout the rest of this document.
+
+* A Crowbar admin node with the Crowbar version you are developing against
```
matrixik: Would be good to mention here to install `git-core` and `patch` tools. New cloud created with `mkcloud` don't have them, maybe it `mkcloud` could install them automatically?

doc/barclamp.md
```diff
@@ -0,0 +1,1224 @@
...
+
+export GIT_AUTHOR_NAME="$C_NAME"
+export GIT_AUTHOR_EMAIL="$C_EMAIL"
+export GIT_COMMITTER_NAME="$C_NAME"
+export GIT_COMMITTER_EMAIL="$C_EMAIL"
+
+alias Cpatch="(cd $BARCLAMP_FORK_PATH; git show | patch -N -p1 --merge -d /opt/dell)"
+
+# Full patch: applies your fork completely against /opt/crowbar
+function Fpatch()
```
matrixik: Anyone else have problem with first run of `Fpatch` on new cloud creating file `/opt/dell/crowbar_framework/barclamps/*.yml` (yes, with `*` in the name)?
`sync_crowbar` fail with this file.

doc/barclamp.md
```diff
@@ -0,0 +1,1224 @@
...
+  {
+  cd $BARCLAMP_FORK_PATH
+  git diff "$COMMIT^" HEAD | patch -N --merge -p1 -d /opt/dell
```
matrixik: When I create new files this is working for me only the first time. Will create files in /opt/dell.
But running it second time it's showing:
```
The next patch would create the file chef/cookbooks/monasca/README.md,
which already exists!  Skipping patch.
1 out of 1 hunk ignored
```
for all new files. So new files are not updated with new content.
It will patch only already existing files:
```
patching file chef/cookbooks/barbican/recipes/api.rb
patching file chef/cookbooks/horizon/recipes/server.rb
```

doc/barclamp.md
```diff
@@ -0,0 +1,1224 @@
...
+# somewhere other than /root/crowbar-openstack)
+
+export BARCLAMP_FORK_PATH=/root/crowbar-openstack
+
+export GIT_AUTHOR_NAME="$C_NAME"
+export GIT_AUTHOR_EMAIL="$C_EMAIL"
+export GIT_COMMITTER_NAME="$C_NAME"
+export GIT_COMMITTER_EMAIL="$C_EMAIL"
+
+alias Cpatch="(cd $BARCLAMP_FORK_PATH; git show | patch -N -p1 --merge -d /opt/dell)"
```
matrixik: Patching sometimes add same content many times to files in `/opt/dell` folder.
Sometimes files in `/opt/dell` end with git merge conflict markers.

doc/barclamp.md
jgrassler: I've had that happen before, but only once. I may have fixed it locally and forgotten to commit it. I'll try to reproduce it on a clean Crowbar node and fix the problem...

doc/barclamp.md
```diff
@@ -0,0 +1,1224 @@
...
+  in `/opt/dell`.
+
+Using patches against `/opt/dell` may seem a bit over-the-top at first since
+one could simply copy files back and forth or use `barclamp_install` on your
+checkout, but it prevents a lot of breakage in cases where `/opt/dell` is out
+of sync with the `crowbar-openstack` repository's master branch or your own
+fork of `crowbar-openstack` is out of sync with upstream.
+
+In the former case copying files from your fork to `/opt/dell` or using
+`barclapm_install` may cause an inconsistent state and errors entirely
```
jgrassler: Fixed.

doc/barclamp.md
jgrassler: It's only supposed to be run once, with Cpatch being run after every new commit later. That _is_ a bit clumsy, though. I think I'll put some more effort into `Fpatch` to so it can do `Cpatch`'s job as well and not make a mess. Stay tuned :-)

doc/barclamp.md
jgrassler: Yes. That would happen with multiple `Fpatch` invocations. See above. I'll try to figure out a way to simplify the whole patching bit.

doc/barclamp.md
aspiers: That doesn't work; maybe you mean `git log --format=%h -n1 master..HEAD`?  And you might need `--reverse` too; I'm not sure whether you want the oldest or newest commit.

jgrassler: > That doesn't work

It does. It yields the oldest commit's short ID, which is what I use for creating the diff further down. Adding your `--format=%H` option would improve it a little in case of a duplicated short ID somewhere, but that's the extent to which it could be said not to work.

But the point is moot anyway: I don't want to rely on the `master` branch being in a usable state or somebody forgetting to switch the reference branch to the appropriate stable branch when doing a backport. It's a little bit of effort to dig up and paste the commit ID manually, but if it does go wrong it's easier to figure out where it went off the rails.

> I'm not sure whether you want the oldest or newest commit.

Good point. "first commit" is a rather ambiguous formulation. Fixed :-)

doc/barclamp.md
jgrassler: I couldn't reproduce it this time around. I got rid of the wildcards, though, so this should no longer occur.

doc/barclamp.md
jgrassler: Forgot to respond to this comment yesterday: good point, thanks. I always do this on autopilot so I forgot mentioning it. I added these two to the prerequisites list now.

doc/barclamp.md
jgrassler: Alright, I replaced `Fpatch` by its idempotent cousin `reposync_crowbar` now (and got rid of `Cpatch` because `reposync_crowbar` will work in that situation too). Worked well for me, but feel free to give it a spin yourself.

doc/barclamp.md
jgrassler: That should be fixed now with `reposync_crowbar` and its patch level record.

doc/barclamp.md
```diff
@@ -0,0 +1,1321 @@
...
+# Insert the path to your barclamp collection checkout here (in case it's
+# somewhere other than /root/crowbar-openstack). This directory must be a git
+# working copy of the crowbar-openstack fork and branch you are working on.
+
+export BARCLAMP_FORK_PATH=/root/crowbar-openstack
+
+export GIT_AUTHOR_NAME="$C_NAME"
+export GIT_AUTHOR_EMAIL="$C_EMAIL"
+export GIT_COMMITTER_NAME="$C_NAME"
+export GIT_COMMITTER_EMAIL="$C_EMAIL"
```
aspiers: Again, this would not be necessary if developing from your native dev environment, but if you are going to do this, it's better to do it via `git config`, so that it applies for all git commands, not just when you are running this script or running git commands in a shell where you forgot to first run these `export` commands.

doc/barclamp.md
```diff
@@ -0,0 +1,1321 @@
...
+    . /opt/dell/patchlevel
+    head=$(cd $BARCLAMP_FORK_PATH; git log --format='%H'| head -n 1)
+    if [ "$head" = "$commit" ]; then
+      echo "/opt/dell is already up to date. If this is not what you " 1>&2
+      echo "expected to see, you may have uncommitted changes" 1>&2
+      echo "in ${BARCLAMP_FORK_PATH}." 1>&2
+      return
+    fi
+  else
+    commit="${COMMIT}^"
```
aspiers: IIUC, this will just end up pointing to `$branch` from above, so I'm not sure why you don't just ask the user to provide that value instead of manually calculating another commit from which you end up with the original answer.  Maybe I'm misunderstanding something - let's mumble quickly.

doc/barclamp.md
```diff
@@ -0,0 +1,1321 @@
...
+  fi
+
+  pushd $BARCLAMP_FORK_PATH > /dev/null
+  cp -a /opt/dell /opt/dell.orig
+  git diff $commit HEAD | patch -N --merge -p1 -d /opt/dell || return
```
aspiers: This feels like a partial reinvention of [this wheel](https://github.com/crowbar/crowbar/blob/master/Guardfile).  I think I understand your reasons for using `patch`, but I disagree with them as explained below. Maybe I'm missing something though.

doc/barclamp.md
```diff
@@ -0,0 +1,1321 @@
...
+
+Using patches against `/opt/dell` may seem a bit over-the-top at first since
+one could simply copy files back and forth or use `barclamp_install` on your
+checkout, but it prevents a lot of breakage in cases where `/opt/dell` is out
+of sync with the `crowbar-openstack` repository's master branch or your own
+fork of `crowbar-openstack` is out of sync with upstream.
+
+In the former case copying files from your fork to `/opt/dell` or using
+`barclamp_install` may cause an inconsistent state and errors entirely
+unrelated to your change.
```
aspiers: If that's the case then I'd argue that the development environment is fundamentally broken, since you're developing and testing an environment which will never appear again, as soon as your changes hit CI.  To be clear: you'd be developing and testing patches which are applied against an old commit, and then submitting them upstream as applied to the latest trunk.  A lot of the time you might get away with this without any major problems, but it still leaves you wide open to introducing bugs which were not caught during development and manual testing of the changes which are often more thorough regarding that particular code area than the CI will be (unless of course the code in question has good test coverage, but even then this makes me uncomfortable).

doc/barclamp.md
```diff
@@ -0,0 +1,1321 @@
...
+
+In the latter case copying files from `/opt/dell` to your fork may re-introduce
+old code that is no longer present upstream. Both should not be an issue for
+new barclamps, but it is better to get used to the `patch(1)` based workflow
+right away since these problems do become a problem when changing an existing
+barclamp later and thus the `patch(1)` based workflow becomes neccessary.
```
aspiers: Again, IMHO this is a broken workflow.  You should first rebase your topic branch so that it matches the `/opt/dell` which you are testing on.

doc/barclamp.md
aspiers: Too late - some of the necessary ingredients already got deleted: https://github.com/crowbar/crowbar-core/pull/1026#issuecomment-275105884

doc/barclamp.md
```diff
@@ -0,0 +1,1321 @@
...
+
+Once they are happy they will approve your pull request. Your pull request will
+need two approvals by Crowbar team members to get merged. Before it gets
+merged, please rebase your pull request so it only consists of a single commit.
+
+#### Adding Test Automation
+
+To include your barclamp in `mkcloud` test automation you would fork the
+[automation](https://github.com/SUSE-Cloud/automation) repository, add your
+barclamp to the list of barclamps enabled by `qa_crobarsetup.sh` and submit a
```
aspiers: `crobar` typo

doc/barclamp.md
```diff
@@ -0,0 +1,1321 @@
...
+[follow-up commit](https://github.com/SUSE-Cloud/automation/commit/0d9d0115fcb8eb236e16effed2e5e5a354fb142a) 
+as not all Barbican packages were available at the time of the first commit.
+
+In the simplest case you should be able to combine something analogous to this
+commit into two pull request. For less than simple scenarios (e.g. if your
+barclamp needs VM side infrastructure such as additional block devices in
+place) you will have to add additional setup code to 
+[mkcloud](https://github.com/SUSE-Cloud/automation/blob/master/scripts/mkcloud)
+and/or
+[qa_crowbarsetup.sh](https://github.com/SUSE-Cloud/automation/blob/master/scripts/qa_crobarsetup.sh)
```
aspiers: `crobar` typo

doc/barclamp.md
```diff
@@ -0,0 +1,1321 @@
...
+  for barclamp in $BARCLAMPS
+    do
+    knife cookbook upload -o /opt/dell/chef/cookbooks/ $barclamp
+    done
+  systemctl restart crowbar
+  }
+
+```
+
+We will make heavy use of the `reposync_crowbar` command thus defined in the
```
aspiers: Please can you take a look at https://github.com/crowbar/crowbar/blob/master/Guardfile and http://publishwith.me/crowbar-tree-merging and then consider whether a more unified approach might be possible?  (I'm not claiming it would be easy, but at least it would tackle the fundamental problem that the git repo structure doesn't match the deployed structure.)  If you'll forgive the history lesson, a long time ago in a universe far away, someone started to write a shell script a bit like this, with the goal of making Crowbar development smoother.  It grew into an 8,000-line monstrosity which could have been avoided by asking some more fundamental questions about development environments and workflows.  I'm not saying there's a similar risk here, but I think the KISS principle still applies :)  Would be very interested in your thoughts.

doc/barclamp.md
```diff
@@ -0,0 +1,1321 @@
...
+```
+reposync_crowbar   # Fix any merge conflicts reported by this command before proceeding
+```
+
+If you navigate to your Crowbar Admin Node's web interface now and select
+*OpenStack* from the barclamp menu, you should see your new barclamp listed. If
+it doesn't appear or you got an error message from the `sync_crowbar` command
+you will now have to start debugging. Common errors at this stage are mostly
+bad file names, or problems in `mybarclamp.yml`.
+
```
aspiers: I think there's a missing step here, or somewhere around here.  You cover local testing via some mechanism for quickly transferring changes from a git repo into `/opt/dell` (and you suggest `reposync_crowbar` which is one possibility but I have also mentioned others).  This speed of transferral is of course important to keep the hack/test cycle efficient.  So far so good.  But when the changes are stabilising, it then makes sense to sacrifice some development speed for higher quality testing, and in general it's more reliable to test by deploying rpms, which get the test deployment closer to a real world deployment.

The downside of testing via rpm deployments is that it takes a lot longer to build an rpm from a git topic branch and then deploy it to a local test cloud.  But that's exactly why [I automated this process down to a one-line command!](https://github.com/openSUSE/pack-tools/tree/master/contrib/BS-pkg-testing)  So I would like to see that also covered in this guide.

aspiers:
BTW, once this gets merged, please ensure it gets linked from https://github.com/SUSE/cloud/tree/master/docs - thanks! /cc @nkrinner

doc/barclamp.md
```diff
@@ -0,0 +1,1321 @@
...
+the `Save` button in the barclamp's view you reached at the end of the previous
+step. This will cause the automatically generated proposal to be validated and
+saved to Crowbar's database. If this fails this is usually due to one of the
+following reasons:
+
+* Type mismatches between `template-mybarclamp.json` and
+  `template-mybarclamp.schema`. This is the most common error.
+
+* Logical errors in `crowbar_framework/app/models/mybarclamp_service.rb`. These
+  may most commonly are role constraint mismatches or fields required by the
```
nkrinner: Not a native speaker, but the 'may' can be dropped?

doc/barclamp.md
```diff
@@ -0,0 +1,1321 @@
...
+In the former case copying files from your fork to `/opt/dell` or using
+`barclamp_install` may cause an inconsistent state and errors entirely
+unrelated to your change.
+
+In the latter case copying files from `/opt/dell` to your fork may re-introduce
+old code that is no longer present upstream. Both should not be an issue for
+new barclamps, but it is better to get used to the `patch(1)` based workflow
+right away since these problems do become a problem when changing an existing
+barclamp later and thus the `patch(1)` based workflow becomes neccessary.
+
```
matrixik: Then maybe you could share what is your workflow?

doc/barclamp.md
aspiers: Yep.  "These are most commonly"

dirkmueller:
@jgrassler have a chance to address the comments? maybe we could merge it then.

jgrassler:
I'm dashing back and forth between a bunch of right now, so probably not today. I'll just make the time to address them tomorrow, though. I want to get this moving along, too...

doc/barclamp.md
```diff
@@ -0,0 +1,1321 @@
...
+end, you will need to re-run mkcloud after you have created your cloud:
+
+```
+nodenumberlonelynode=1 mkcloud setuplonelynodes crowbar_register
+
+```
+
+## Development Environment
+
+Most of your development will take place on your Crowbar node (log in as root).
```
jgrassler: Yes. In fact this is what I am counting on. It will provide a uniform environment where editor, configuration files, ssh keys (or the lack thereof are known quantities).

doc/barclamp.md
```diff
@@ -0,0 +1,1321 @@
...
+You will find the currently running Crowbar code (both Rails application
+components and chef recipes) in the `/opt/dell` directory. We will provide
+detailed locations for all important components of a Barclamp in the _Basic
+Implementation_ section below.
+
+In order to track changes you make we recommend you keep a checkout of your
+`crowbar-openstack` fork on the Crowbar node and apply the changes you make in
+there
```
jgrassler: It would be nicer and developers are free to work that way. As a documentation writer that presents me with a moving target, though.

doc/barclamp.md
```diff
@@ -0,0 +1,1321 @@
...
+comments):
+
+```
+# Replace "Crowbar Committer" by your name as it should appear in commit
+# messages.
+C_NAME="Crowbar Committer"
+
+# Replace "email@address.invalid" by your email address as it should appear in
+# commit messages.
+C_EMAIL="email@address.invalid"
```
jgrassler: Yes. This makes collaboration that much harder, though. One of the main aims of this approach is providing an environment everyone can live with and use in a collaboration scenario on a remote Crowbar node.

doc/barclamp.md
```diff
@@ -0,0 +1,1321 @@
...
+# where $branch is the upstream branch you are developing against such as
+# `master` or `stable/3.0`, should give you a list of original commits. This
+# assumes $branch has not been improperly cherry-picked into or rebased in your
+# fork, though. It also assumes that you are not working on your fork's
+# `master` branch (you shouldn't, but people have been known to do this
+# anyway). Since all of these may occasionally happen, this guide leaves it up
+# to the user to set $COMMIT as they see fit rather than trying to determine
+# its value automatically through the command mentioned above.
+
+export COMMIT="6cefbd80254141a63f4c86df195cd930ac7eb10b"
```
jgrassler: There may be multiple commits in a topic branch. Not everyone uses the OpenStack model of `git commit --amend` that leaves just a single commit in a topic branch. It doesn't make sense without Gerrit or similar heavy CI artillery. Likewise, `master` or `stable/3.0` may be way out of date in a fork (it commonly is). The raw SHA1 guarantees you a diff from your first commit to HEAD on the topic branch. Nothing more, nothing less. That is why I opted for this.

doc/barclamp.md
jgrassler: That is why I put these commands into .bashrc. I wanted to keep them in the environment because then they can be changed in a collaboration scenario without clobbering the hardwired configuration as set by `git config`.

doc/barclamp.md
jgrassler: Yes, of sorts. But given the uncertainties involved with diffing `$branch` against a stale `master` or stale `stable/3.0` I went for calculating the diff this way. This simply formats a range of commits as a patch, regardless of the (possibly stale) state of other branches.

doc/barclamp.md
jgrassler: You may be. Let me try and clarify my intentions behind re-inventing that particular wheel below. Maybe we'll find some common ground that way.

doc/barclamp.md
jgrassler: > Please can you take a look at https://github.com/crowbar/crowbar/blob/master/Guardfile and
http://publishwith.me/crowbar-tree-merging and then consider whether a more unified approach might be possible? (I'm not claiming it would be easy, but at least it would tackle the fundamental problem that the git repo structure doesn't match the deployed structure.)

Not everybody will happy to perform
https://github.com/crowbar/crowbar/blob/master/Guardfile#L20 on their local
machine. But, more importantly, this approach is awkward in a collaboration scenario. That being said I added a link for those who prefer
working locally.

As for http://publishwith.me/crowbar-tree-merging: I'm cautiously optimistic
that we might have a unified solution at some stage, but the reality is that
we do not have a unified solution right now. Nor do I intend to provide one
with that little hack. It's just something a new developer can use for their
first steps (after that they'll build something of their own anyway, and _that_
may end up they unified solution we are looking for).

> If you'll forgive the history lesson, a long time ago in a universe far away, someone started to write a shell script a bit like this, with the goal of making Crowbar development smoother. It grew into an 8,000-line monstrosity which could have been avoided by asking some more fundamental questions about development environments and workflows. I'm not saying there's a similar risk here, but I think the KISS principle still applies :) Would be very interested in your thoughts.

I bore the KISS principle in mind as I came up with this :-)

While the script grew some tentacles, still I tried to limit these as far as
possible. Also, I put it in here on purpose: I consider it more
documentation/inspiration than code to be run as-is since it shows what needs
to be done to synchronize the state of one's `crowbar-openstack` fork with a
running Crowbar instance. I only intend to change it if these synchronization
needs change. It is definitely not supposed to solve everybody's development
environment problems. Where that lofty goal is concerned its only aspiration is
showing people what challenges they face if they chose to work on that solution.

doc/barclamp.md
jgrassler: Crowbar changes will take a while to percolate through CI and end up in the RPMs a development cloud is built from. Also, gating breakage happens and may hinder that percolation, sometimes for days on end. That way even a freshly built cloud can be out of date. Of course one should still give a patch a manual test run against a freshly built cloud, but the RPMs you get are occasionally behind Crowbar's `master` branch, and for these situations it is nice to have a tool that enables you to still continue development.

doc/barclamp.md
jgrassler: Oh. I didn't know that one :-)

Thanks!

I included a link to that as well.

doc/barclamp.md
jgrassler: Thanks, fixed!

doc/barclamp.md
```diff
@@ -0,0 +1,1321 @@
...
+
+  * `patch`
+  * `git-core`
+
+Providing these is out of scope for this document. We assume you are able to
+fork the appropriate Barclamp collection. The machines can be provided through
+manual installation or by using an automated setup tool such as
+[mkcloud](https://github.com/SUSE-Cloud/automation/blob/master/scripts/mkcloud).
+Depending on the individual situation you can use either the `plain` or the
+`installonly` `mkcloud` step. The former will install the nodes and deploy the
```
jgrassler: Thanks, fixed!

doc/barclamp.md
```diff
@@ -0,0 +1,1321 @@
...
+standard set of barclamps in their default configuration, while the latter will
+only install the nodes. Use `plain` if your barclamp does not need to be
+deployed before other barclamps and if you do not modify other barclamps.
+Otherwise use `installonly` and deploy barclamps bit by bit through the Crowbar
```
jgrassler: Thanks, fixed!

doc/barclamp.md
```diff
@@ -0,0 +1,1321 @@
...
+
+## Barclamp Component Overview
+
+A barclamp consists of two main components:
+
+* A [chef cookbook](https://docs.chef.io/cookbooks.html) which takes care of
+  actually deploying your service.
+* Some glue code to tie your chef cookbook into the Crowbar Web UI. This glue
+  code will parametrize your chef cookbook and apply its component roles to the
+  nodes it orchestrates.
```
jgrassler: Ok, fixed

doc/barclamp.md
```diff
@@ -0,0 +1,1321 @@
...
+          "cluster" => true,   # Whether this role can be clustered (important for HA later on).
+          "admin" => false,    # Whether this role can be assigned to the Crowbar admin node.
+          "exclude_platform" => {
+            "suse" => "< 12.1",  # Do not deploy on SUSE version 12.1 and lower.
+            "windows" => "/.*/"  # Do not deploy on any Windows platforms.
+          }
+        },
+         "mybarclamp-compute" => {
+          "unique" => false,
+          "count" => -1,   # Can be deployed to an arbitrary number of Compute nodes
```
jgrassler: Yes...I wanted to have a concrete usage example as well, in case somebody missed the comment above.

jgrassler:
Alright, I think I addressed/discussed all current comments now. Let's see if we can reach a consensus on this...

Also, thank you for your suggestion on the [Lx:Ly style links](https://github.com/jgrassler/crowbar/pull/1), @sjamgade! I integrated it, but I had already rebased when I thought of the still open pull request.

doc/barclamp.md
aspiers: OK, sounds good!

doc/barclamp.md
aspiers: Cool, thanks :)

doc/barclamp.md
aspiers: That kind of misses the point.  You're right that the *other* RPMs can lag behind `master`, but I'm talking about the RPM which you're actually hacking on in this context, which is presumably `crowbar-openstack`.  My point was that you're generating a patch from the topic branch which is based off the (merge-)base of the topic branch, and then applying that patch against whatever revision of `crowbar-openstack` happens to be deployed in `/opt/dell`.  That's akin to a three-way merge, and is hence prone to not only merge conflicts (which are OK, because they're visible breakage which can be fixed) but also invisible breakage due to incompatibilities between the two base versions in the merge.

Having said all that, Dirk reminded me earlier today that in this particular context of creating a new barclamp, it is far less likely that this kind of breakage would occur, because all we're doing really is adding new files.  So it would only break if there was some incompatible API change in the barclamp framework, and that's fairly unlikely most of the time.

So in summary, I'm not too worried about the downsides of your approach *in this context*, although I'd recommend thinking carefully before using that in other "normal" hacking contexts.

doc/barclamp.md
aspiers: I'm 85% sure that this comment is misleading and that at least one if not two or all three of these are in use.

doc/barclamp.md
jgrassler: They have an identical copy of the script, since it's in the .bashrc. B can change the environment variables ad-hoc in just one terminal if the name/email address for A are named in the .bashrc. This ad-hoc approach won't change the persistent configuration, which prevents A from being surprised at suddenly committing as B.

I know that the same could be achieved by using `git config --global` and using environment variables only for an override. I opted for using the environment variable approach for the persistent configuration as well since it's a tiny little bit faster, not involving a file on disk and, more importantly, documents how to do the ad-hoc change for user B in a copy & paste friendly manner.

doc/barclamp.md
jgrassler: "lowest common denominator" captures it pretty well, yes :-)

For any 10 developers you'll have about 25 vastly differing local development environments. Instead of trying to come up with something all of them are happy with I wanted to accomplish two things:

* Provide something anybody can use even if their workstation accidently got an anvil dropped on it an hour ago and now they have a new machine fresh from IT with none of their customized personal environment on it.
* Provide something that shows them all the bits they need to touch in a compact, easy to understand manner so they can roll their own development environment if they need to or fix the Guardfile themselves if it bitrots.

aspiers:
Just discussed this with @jgrassler and we agreed he will:

- move the info about the option of using a local develop environment so that it appears before the detailed instructions on setting up a dev environment on the admin server
- drop all the `$COMMIT` stuff and instead use the 3-dot form of `git diff` to automatically generate the patch from the merge-base of the topic branch
- fix the issue with the `cp` to a `.bak` directory

After that, I'll be an enthusiastic +1 :-)

doc/barclamp.md
```diff
@@ -85,24 +124,14 @@ nodenumberlonelynode=1 mkcloud setuplonelynodes crowbar_register
...
-
 ### Preparations for Crowbar Node
 
 These are recommendations, but we'll assume you followed them for the remainder
 of this document since this shortens the commands in our examples. First of
 all, clone your `crowbar-openstack` fork to `/root/crowbar-openstack` on your
-Crowbar node and switch to your development branch.
+Crowbar node and switch to your development branch. Note, that the code below
+assumes you are working on a dedicated topic branch. It will break if you are
+doing development on your fork's `master` branch.
```
aspiers: Yes, you could even be more forceful and say "DON'T WORK ON master!" :)

doc/barclamp.md
```diff
@@ -168,14 +181,35 @@ function reposync_crowbar()
...
+  else
+    echo "Found a previous intermediate backup of /opt/dell in /opt/dell.bak" 1>&2
+    echo "This indicates a failed reposync_crowbar run." 1>&2
+    echo "Please restore /opt/dell to a known good state from either of" 1>&2
+    echo "    /opt/dell.bak" 1>&2
+    echo " or /opt/dell.orig" 1>&2
+    echo " and remove /opt/dell/back.bak once you have done so." 1>&2
+    return 1
+  fi
+  if [ -n "$commit" ]; then  # only apply commits added since the last reposync_crowbar run
```
aspiers: Might be worth a comment that this can only get set via `/opt/dell/patchlevel` now.

doc/barclamp.md
jgrassler: Done.

doc/barclamp.md
jgrassler: Done.

PR #2296, opened 2017-01-02
upgrade: set status to failed when admin upgrade failed (bsc#1015312)

MaximilianMeister:
https://bugzilla.suse.com/show_bug.cgi?id=1015312

### DIFF
--- a/scripts/upgrade_admin_server.sh
+++ b/scripts/upgrade_admin_server.sh
@@ -65,8 +65,20 @@ upgrade_admin_server()
     if [ $ret != 0 ]; then
         # In the failed case, crowbar should tell user to check zypper logs,
         # fix the errors and continue admin server manually
-        echo "zypper dist-upgrade has failed with $ret, check zypper logs"
+        local errmsg="zypper dist-upgrade has failed with $ret, check zypper logs"
+        echo $errmsg
         echo "$ret" > $UPGRADEDIR/admin-server-upgrade-failed
+        # The status of the upgrade needs to be set in case of an error
+        # so we can just use the upgrade_status library to set the status to "failed"
+        # otherwise the status of the admin_upgrade would stay at "running" which would prevent
+        # continuation of the upgrade even if the admin upgrade got fixed manually
+        ruby -e "
+            require 'logger'
+            require '/opt/dell/crowbar_framework/lib/crowbar/upgrade_status'
+            ::Crowbar::UpgradeStatus.new(Logger.new(STDOUT)).end_step(
+              false,
+              '$errmsg'
+            )"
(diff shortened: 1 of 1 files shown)


MaximilianMeister:
[Bug 1015312 - [upgrade][api] If admin upgrade zypper fails, api will get stuck on admin_upgrade: running](https://trello.com/c/5ZxaWYBT/228-bug-1015312-upgrade-api-if-admin-upgrade-zypper-fails-api-will-get-stuck-on-admin-upgrade-running)

scripts/upgrade_admin_server.sh
```diff
@@ -65,8 +65,16 @@ upgrade_admin_server()
     if [ $ret != 0 ]; then
         # In the failed case, crowbar should tell user to check zypper logs,
         # fix the errors and continue admin server manually
-        echo "zypper dist-upgrade has failed with $ret, check zypper logs"
+        local errmsg="zypper dist-upgrade has failed with $ret, check zypper logs"
+        echo $errmsg
         echo "$ret" > $UPGRADEDIR/admin-server-upgrade-failed
+        ruby -e "
```
jsuchome: Please add a comment here what the following action is about

MaximilianMeister: added

scripts/upgrade_admin_server.sh
```diff
@@ -65,8 +65,20 @@ upgrade_admin_server()
...
         # fix the errors and continue admin server manually
-        echo "zypper dist-upgrade has failed with $ret, check zypper logs"
+        local errmsg="zypper dist-upgrade has failed with $ret, check zypper logs"
+        echo $errmsg
         echo "$ret" > $UPGRADEDIR/admin-server-upgrade-failed
+        # The status of the upgrade needs to be set in case of an error
+        # so we can just use the upgrade_status library to set the status to "failed"
+        # otherwise the status of the admin_upgrade would stay at "running" which would prevent
+        # continuation of the upgrade even if the admin upgrade got fixed manually
+        ruby -e "
```
skazi0: maybe we should create a small cli script for updating the status from shell scripts?

MaximilianMeister: @jsuchome do we need that for the node upgrade somehow?

jsuchome: AFAIK, we do not need it for any other case than this. I assume @skazi0 is suggesting it to make things clearer

skazi0: I thought there will be some other uses for such script. If this is the only one we have currently, I'm fine with an embedded implementation.

PR #2297, opened 2017-01-02
upgrade: Remove old chef-created crowbar.service unit

rhafer:
It's part of the crowbar-core package nowadays


rhafer:
Requires: https://github.com/crowbar/crowbar-core/pull/941

PR #2301, opened 2017-01-04
install-chef-suse: Create admin node in DB

rsalevsky:
We expect that the admin node already exists in the database and the
mechanism for the nodes doesn't catch this case.

Needs to go together with https://github.com/crowbar/crowbar-core/pull/920


nicolasbock:
Gating is failing with
~~~
uninitialized constant Node (NameError)
~~~

PR #2303, opened 2017-01-16
install-chef-suse: Run db:seed after database cleanup

rsalevsky:
Part of https://github.com/crowbar/crowbar-core/pull/920

### DIFF
--- a/scripts/install-chef-suse.sh
+++ b/scripts/install-chef-suse.sh
@@ -291,7 +291,7 @@ function reset_crowbar()
 {
     rm -f $crowbar_install_dir/crowbar-install-failed
     rm -f $installation_steps
-    su -s /bin/sh - crowbar sh -c "cd /opt/dell/crowbar_framework && RAILS_ENV=production bin/rake db:cleanup"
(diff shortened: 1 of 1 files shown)


scripts/install-chef-suse.sh
```diff
@@ -291,7 +291,7 @@ function reset_crowbar()
 {
     rm -f $crowbar_install_dir/crowbar-install-failed
     rm -f $installation_steps
-    su -s /bin/sh - crowbar sh -c "cd /opt/dell/crowbar_framework && RAILS_ENV=production bin/rake db:cleanup"
+    su -s /bin/sh - crowbar sh -c "cd /opt/dell/crowbar_framework && RAILS_ENV=production bin/rake db:cleanup db:seed"
```
felixsch: Why is a db:seed needed here? db:cleanup looks like a task which basically migrate the database. Did I get something wrong?

rsalevsky: `db:cleanup` resets the complete database which means that we loose all the data including the default values. ATM we don't have any default values but after the switch to the Node in the DB the admin node will be such a default value.

felixsch: Ok didn't know that the migrator also removes data.

PR #2304, opened 2017-01-27
Added back checks for SLES12SP2 based repositories, to be used

jsuchome:
when SLES12SP2 is the default. They were temporary removed in
commit bf5a372a4770b0bb4853a7cc842c20c322fb2695

(cherry picked from commit 9240ed580c5c2ad028af70248276636cacaf36fb)

### DIFF
--- a/scripts/install-chef-suse.sh
+++ b/scripts/install-chef-suse.sh
@@ -549,7 +549,7 @@ for arch in $product_arches; do
             $MEDIA \
             $sp2mediasum
 
-        if ! is_ses; then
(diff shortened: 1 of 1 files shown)


dirkmueller:
@jsuchome could you check this? PR gating seems to fail

jsuchome:
@dirkmueller should the check for Cloud repo be ignored if arch != x86_64? 
I think it fails while calling `check_media_content Cloud /srv/tftpboot/suse-12.2/aarch64/repos/Cloud`

PR #2305, opened 2017-02-03
add example backup cronjob

bmwiedemann:
because otherwise it is not trivial for admins to do
based on onadmin_crowbarbackup function

### DIFF
--- a/scripts/crowbar-backup-cronjob
+++ b/scripts/crowbar-backup-cronjob
@@ -1,0 +1,6 @@
+#!/bin/sh
+# example backup script to be added in /etc/cron.daily/
+btarballname=crowbar-$(date +%F)
+p="--password=crowbar"
+crowbarctl backup create $btarballname $p && \
+echo "Created backup in /var/lib/crowbar/backup/$btarballname.tar.gz"


scripts/crowbar-backup-cronjob
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+# example backup script to be added in /etc/cron.daily/
+dir=/tmp/backup
+btarballname=crowbar-$(date +%F)
+p="--password=crowbar"
+mkdir -p $dir
```
dirkmueller: Mkdir -p on a hardcoded tmp dir name..

scripts/crowbar-backup-cronjob
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+# example backup script to be added in /etc/cron.daily/
+dir=/tmp/backup
```
dirkmueller: Why do we need to cd into that directory? If you need a scrstch space, better use one that is only root writeable

bmwiedemann:
updated to /root/backup to be a better (more secure) example

scripts/crowbar-backup-cronjob
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+# example backup script to be added in /etc/cron.daily/
+dir=/root/backup
+btarballname=crowbar-$(date +%F)
+p="--password=crowbar"
+mkdir -p $dir
+cd $dir
+crowbarctl backup create $btarballname $p && \
+bid=`crowbarctl backup  list --plain $p | grep ${btarballname} | cut -d" " -f1` && \
+crowbarctl backup download $bid $p
```
rsalevsky: Why should we store the backup twice on the admin node?

bmwiedemann: because crowbarctl did not tell that it wrote it to `/var/lib/crowbar/backup/`. will update the script.

bmwiedemann:
Max suggested, that it could also use an environment variable or a .crowbarrc for passing the password, but that seems undocumented.

PR #2306, opened 2017-02-10
upgrade: Report the admin server upgrade error if knife has failed

jsuchome:


jsuchome:
[Changing platform during admin server upgrade failed](https://trello.com/c/wXAOM9Hi/309-changing-platform-during-admin-server-upgrade-failed)

scripts/upgrade_admin_server.sh
```diff
@@ -91,8 +99,13 @@ n.target_platform = 'suse-12.2'
 n.provisioner.default_os = 'suse-12.2'
 n.save"
 
-    # Signalize that the upgrade correctly ended
-    echo "12.2" >> $UPGRADEDIR/admin-server-upgraded-ok
+    ret=$?
+    if [ $ret != 0 ]; then
+        report_failure($ret, "Setting the platform to suse-12.2 has failed.")
```
rhafer: Not sure if it makes a bit difference here. But shouldn't we acutally exit here, instead of just skipping the ```admin-server-upgraded-ok``` step

jsuchome: Yes, I was thinking about that, and I'm not sure. It may actually make sense to run the rest of the script anyway, and user could try to just fix these os values on the upgraded server.

I'm not sure if failing on Cloud6 side makes sense, as the crowbar has been already upgraded - I think restarting the whole step will not be possible. (This is related also to the other possible failure actually)

jsuchome: That said ... if users do not restart the step using the progress library, it stays as failed and they cannot continue. Sigh.

jsuchome:
> agree with @rhafer that exiting on that check may be is the best option.

This is really complicated. I think that once packages are upgraded, we need to let the admin server reboot and handle possible failures on Cloud7 side. Otherwise I'm not sure what the UI should tell the user if it detects the failure during this step: upgraded crowbar has already Cloud7 sources...

rhafer:
> This is really complicated. I think that once packages are upgraded, we need to let the admin server reboot and handle possible failures on Cloud7 side.

In that case wouldn't it make sense to just run the knife command before upgrading the packages? (i.e. after stopping chef-client)

jsuchome:
> In that case wouldn't it make sense to just run the knife command before upgrading the packages? (i.e. 
> after stopping chef-client)

That might work actually... I have to test it

PR #2316, opened 2017-03-31
Fix admin node timezone detection

dirkmueller:


scripts/install-chef-suse.sh
```diff
@@ -911,7 +911,7 @@ fi
 
 # Use current time zone
 (
-    . /etc/sysconfig/clock || : Ignoring missing /etc/sysconfig/clock
+    TIMEZONE=$(LC_ALL=C timedatectl status | grep "zone:" | cut -d: -f2 | cut -d' ' -f2)
```
nicolasbock: Alternatively you could sed this: `timedatectl status | sed -e "/zone:/ s/^.*: \([^ ]\+\).*/\1/; t; d"`...

aspiers: I think `awk` would be more elegant and readable than either of these:

    timedatectl status | awk '/Time zone:/ {print $3}'

nicolasbock: Yes, that is shorter :smile:

dirkmueller: awk is not installed, but I'm happy you found a place to bikeshed on. we can use the sed solution, I don't really care

aspiers: I don't really care about how it is implemented in this particular case either.  You could choose to see it as bikeshedding, or you could choose to see it as sharing knowledge in the hope that it's useful and can help produce better coding practices in the future.  Personally I find the latter type of collaboration more positive and enjoyable.  I mentioned this approach because `awk` seems generally underused in these kinds of situations, even though I find it can often get the job done more cleanly than standard combinations of `grep`, `cut`, `sed` etc., so I think it deserves some visibility.  If you already knew that then great.  I'm not trying to ram it down anyone's throat though.

dirkmueller: @aspiers sorry for being the target of my frustration here, I didn't mean it that way. I agree the awk suggestion is visually most appealing, but there are a million ways of achieving the same thing here (including writing a erlang program for extracting the time zone), its just not relevant for me in the level of problems that I currently have to deal with.

scripts/install-chef-suse.sh
```diff
@@ -911,7 +911,7 @@ fi
 
 # Use current time zone
 (
-    . /etc/sysconfig/clock || : Ignoring missing /etc/sysconfig/clock
+    TIMEZONE=$(LC_ALL=C timedatectl status | sed -e "/Time zone:/ s/^.*: \([^ ]\+\).*/\1/; t; d")
```
cmurphy: On SLES 12-SP1 `timedatectl status` prints out "Timezone" with no space.

s-t-e-v-e-n-k: It's slightly more complex than that, sadly.

Cloud 6, so SP1:

root@crowbar:~ # LC_ALL=C timedatectl status | sed -e "/Time zone:/ s/^.*: \([^ ]\+\).*/\1/; t; d"
root@crowbar:~ # LC_ALL=C timedatectl status | sed -e "/Time \?zone:/ s/^.*: \([^ ]\+\).*/\1/; t; d"
UTC

Cloud 7, so SP2:

root@crowbar:~ # LC_ALL=C timedatectl status | sed -e "/Time zone:/ s/^.*: \([^ ]\+\).*/\1/; t; d"
Etc/UTC
root@crowbar:~ # LC_ALL=C timedatectl status | sed -e "/Time \?zone:/ s/^.*: \([^ ]\+\).*/\1/; t; d"
Etc/UTC

So not only did the field change between SP1 and SP2, but the output as well.

rsalevsky: I don't see an issue with SP1:
master = Cloud7 = SP2
stable/3.0 = Cloud6 = SP1

aspiers:
Ideally we want it on Cloud 6 too; I guess @rsalevsky's point was that we can do a slightly different PR against `stable/3.0` to handle that, rather than just cherry-picking this directly.

dirkmueller:
we could just go back to my original pull request :)

dirkmueller:
@nicolasbock @cmurphy please recheck

PR #2319, opened 2017-04-12
upgrade: Make sure latest chef data are uploaded (bsc#1033160)

jsuchome:
Normally packages should take care of it, but for the rare case of chef-server
not being available during the upgrade, better make sure and do it explicitely.

### DIFF
--- a/scripts/upgrade_admin_server.sh
+++ b/scripts/upgrade_admin_server.sh
@@ -116,6 +116,13 @@ n.save"
     rm -f /etc/systemd/system/crowbar.service
     systemctl daemon-reload
 
+    # Make sure latest chef data are uploaded
+    # Normally packages should take care of it, but for the rare case of chef-server
+    # not being available during the upgrade, better make sure and do it explicitely.
+    systemctl restart chef-server
+    pushd /opt/dell/crowbar_framework
+    sudo -u crowbar RAILS_ENV=production bin/rake chef:upload:all
+
     # cleanup upgrading indication
     # technically the upgrade is not done yet but it has to be
     # done before the reboot


scripts/upgrade_admin_server.sh
```diff
@@ -116,6 +116,13 @@ n.save"
     rm -f /etc/systemd/system/crowbar.service
     systemctl daemon-reload
 
+    # Make sure latest chef data are uploaded
+    # Normally packages should take care of it, but for the rare case of chef-server
+    # not being available during the upgrade, better make sure and do it explicitely.
+    rcchef-server start
```
rsalevsky: I would use `systemctl restart chef-server`.

jsuchome:
[Bug 1033160: chef data not uploaded during the upgrade](https://trello.com/c/WCen6FLw/237-bug-1033160-chef-data-not-uploaded-during-the-upgrade)

jsuchome:
@rsalevsky adapted and tested

scripts/upgrade_admin_server.sh
```diff
@@ -116,6 +116,13 @@ n.save"
     rm -f /etc/systemd/system/crowbar.service
     systemctl daemon-reload
 
+    # Make sure latest chef data are uploaded
+    # Normally packages should take care of it, but for the rare case of chef-server
+    # not being available during the upgrade, better make sure and do it explicitely.
+    systemctl restart chef-server
+    pushd /opt/dell/crowbar_framework
+    sudo -u crowbar RAILS_ENV=production bin/rake chef:upload:all
```
Itxaka: Will this work properly if the chef-server is still booting? No idea how long it takes to restart it and if this command will retry properly?

jsuchome: Hm, do you think we should wait until it is started?

Just a reminder - this is already a backup solution, because normally chef data are uploaded already during package installation and chef-server should not be stopped at that time.

Itxaka: I would be happier if the rake command was the one waiting for the chef server to be available instead of us monkey patching it here.

Not a blocker at all, just a question and a note for any bugs that may arise from this :)

jsuchome:
@rsalevsky do you know why the testbuild keeps failing with `undefined method `keys' for nil:NilClass` ? That does not seem at all related to this PR. Do we have broken crowbar package?

PR #2326, opened 2017-06-14
Switch to SLES12SP3 which is the base for SOC8.

jsuchome:
This should go in after we create 4.0 branch for SOC7

### DIFF
--- a/scripts/install-chef-suse.sh
+++ b/scripts/install-chef-suse.sh
@@ -519,8 +519,8 @@ else
 fi
 
 for arch in $product_arches; do
-    # Checks for SLE12 SP2 medias
-    MEDIA=/srv/tftpboot/suse-12.2/$arch/install
+    # Checks for SLE12 SP3 media
+    MEDIA=/srv/tftpboot/suse-12.3/$arch/install
 
     # Only x86_64 is truly mandatory; other architectures are only checked
     # if they exist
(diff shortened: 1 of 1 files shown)


jsuchome:
There's a conflict in `grub2` package versions.

scripts/install-chef-suse.sh
```diff
@@ -519,8 +519,8 @@ else
 fi
 
 for arch in $product_arches; do
-    # Checks for SLE12 SP2 medias
-    MEDIA=/srv/tftpboot/suse-12.2/$arch/install
+    # Checks for SLE12 SP3 medias
```
aplanas: nit: media?

scripts/install-chef-suse.sh
```diff
@@ -534,25 +534,25 @@ for arch in $product_arches; do
         echo "Detected SUSE OpenStack Cloud Deps media."
     else
 
-        sp2mediasum="a1f512c3244c8e75da0f47d2d0752064"
+        sp3mediasum="59280f20e0bee15615b598a3ea3f027e"
```
dirkmueller: are we sure that those media sums don't change? I would assume we have to comment out the check for now while SP3 isn't released yet..

jsuchome: Yep, this is the sum of current state of medias. I'll comment it.

PR #2327, opened 2017-06-15
[stable/3.0] Fix admin node timezone detection

dirkmueller:
(cherry picked from commit b840885f011554f5890163242eb8c758df0c8ac0)

### DIFF
--- a/scripts/install-chef-suse.sh
+++ b/scripts/install-chef-suse.sh
@@ -884,7 +884,7 @@ fi
 
 # Use current time zone
 (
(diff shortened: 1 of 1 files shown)


scripts/install-chef-suse.sh
```diff
@@ -884,7 +884,7 @@ fi
 
 # Use current time zone
 (
-    . /etc/sysconfig/clock || : Ignoring missing /etc/sysconfig/clock
+    local TIMEZONE=$(LC_ALL=C timedatectl status | grep "zone:" | cut -d: -f2 | cut -d' ' -f2)
```
jsuchome: Drop the `local`

PR #2329, opened 2017-06-16
Do not use local outside of function

dirkmueller:
instcrowbar fails with "local: can only be used in a function"

(cherry picked from commit 6efd9ce76504193697f56eca0493843e0896737c)


dirkmueller:
argh

PR #2332, opened 2017-07-28
Wait longer for DHCP to kick in

dirkmueller:
Apparently in some cases spanning tree blocks dhcp responses for
up to 50s, so wait 60s.


bmwiedemann:
still not sure, if it actually helps. I have seen runs where a ifup took 60 seconds and network still was not up.
The first 35 seconds the network is not transmitting anything. Then there might be some exponential backoff algorithm where it does not send one dhcp request per second or something else I do not yet understand.

nicolasbock:
You mean the DHCP of the ifup took that long? Shouldn't it send out a DHCPREQUEST right away?

bmwiedemann:
It probably does, but due to STP, nothing gets through for 35 seconds... or maybe my netwait=60 parameter made it worse...

PR #2333, opened 2017-08-16
Raise default timeout even further for sleshammer

dirkmueller:
The 60s doesn't seem to be enough for some environments, and there
is no harm in waiting a little bit longer.

### DIFF
--- a/sledgehammer/start-up.sh
+++ b/sledgehammer/start-up.sh
@@ -103,7 +103,7 @@ if ! suse_ver 12; then
 else
     if [ "$BOOTDEV" != "eth0" ]; then
         mv /etc/sysconfig/network/ifcfg-eth0 /etc/sysconfig/network/ifcfg-$BOOTDEV
-        wicked ifup --timeout ${WAIT_FOR_INTERFACES:-60} $BOOTDEV
+        wicked ifup --timeout ${WAIT_FOR_INTERFACES:-120} $BOOTDEV
     fi
 fi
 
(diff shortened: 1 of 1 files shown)


jsuchome:
Would it make sense to have it configurable?

sledgehammer/start-up.sh
```diff
@@ -103,7 +103,7 @@ if ! suse_ver 12; then
 else
     if [ "$BOOTDEV" != "eth0" ]; then
         mv /etc/sysconfig/network/ifcfg-eth0 /etc/sysconfig/network/ifcfg-$BOOTDEV
-        wicked ifup --timeout ${WAIT_FOR_INTERFACES:-60} $BOOTDEV
+        wicked ifup --timeout ${WAIT_FOR_INTERFACES:-120} $BOOTDEV
```
nicolasbock: Didn't you just recently raise this timeout? Do you have a suspicion why this is taking so long?

dirkmueller: @nicolasbock yes. the issue is that in the hardware environment it takes between 59 and 63s  to get network. so 60 was at the border and it was flakey.

dirkmueller:
@jsuchome well, somehow it should be configurable (its an environment variable, it only uses 120 when it isn't set already). I don't know exactly how to inject env variables into the sleshammer image..

aspiers:
> I don't know exactly how to inject env variables into the sleshammer image.

Kernel command-line maybe? Just an idea, might be stupid.

sledgehammer/start-up.sh
nicolasbock: Thanks for the explanation!

nicolasbock: Thanks for the explanation!

toabctl:
Backport is https://github.com/crowbar/crowbar/pull/2334

PR #2335, opened 2017-09-19
rubocop: Set max BlockLength to 40 and ignore spec files

rsalevsky:
The default block lenght is too short for crowbar and leads to many
false positives. Also by nature spec files have longer blocks so let's
ignore them.

### DIFF
--- a/.hound.ruby.yml
+++ b/.hound.ruby.yml
@@ -36,3 +36,8 @@ Metrics/ClassLength:
 
 Metrics/ModuleLength:
   Max: 500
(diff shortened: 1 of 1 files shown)


Itxaka:
Late but thank you @rsalevsky! should we backport this to 4.0 + crowbar-client + crowbar-init?

rsalevsky:
@Itxaka If I understand rupocop then it will always use the configs from here (master) if the other repo doesn't contain any config files.

PR #2336, opened 2017-10-24
Use crowbarctl to activate repositories

cmurphy:
In order to make dealing with an HTTPS crowbar server easier and just to
avoid ugly curl commands, replace the curl call in the install script
with the equivalent crowbarctl call.

https://trello.com/c/2SwQTBg1

### DIFF
--- a/scripts/install-chef-suse.sh
+++ b/scripts/install-chef-suse.sh
@@ -1166,7 +1166,7 @@ touch $crowbar_install_dir/crowbar-installed-ok
 rm -f $crowbar_install_dir/crowbar_installing
 
 # activate provisioner repos
-curl -X POST http://localhost:3000/utils/repositories/activate_all.json
+crowbarctl repository activate-all
 
 kill_spinner
 


scripts/install-chef-suse.sh
```diff
@@ -1166,7 +1166,7 @@ touch $crowbar_install_dir/crowbar-installed-ok
 rm -f $crowbar_install_dir/crowbar_installing
 
 # activate provisioner repos
-curl -X POST http://localhost:3000/utils/repositories/activate_all.json
+crowbarctl repository activate-all
```
cmurphy: Someone sanity-check me that this will be installed on non-mkcloud envs at this point?

rsalevsky: To be sure I would add it to Devel:Cloud:8/patterns-cloud.

cmurphy: This turned out not to be necessary so this is ready to be re-reviewed

vuntz:
Does this need a backport to stable/4.0?

PR #2337, opened 2017-10-27
Added script for setting up chef cluster

skazi0:
Moved from https://github.com/crowbar/crowbar-core/pull/1388

**Why is this change necessary?**
There's currently no "automatic" or "documented" way of setting up chef-server cluster in SOC. Having multiple chef-server instances running at the same time to serve chef-client requests improves performance of large scale installations where single chef-server can be a bottle-neck and can be easily saturated by many concurrent requests from chef-clients.

**How does it address the issue?**
The script can be used to enable/disable multi-worker chef cluster with apache load balancer.

**Is there additional information worth sharing like links to a Trello
card, bug references, testing advice or dependencies to other pull
requests?**
https://trello.com/c/kQ1eLgYY/122-evaluate-performance-of-admin-node-on-scale-cloud

**Usage**
`/opt/dell/bin/chef-server-cluster enable` will configure and start chef cluster with default number of workers based on number of CPUs. If there was some `/etc/sysconfig/chef-server`, the old file will be backed up as `/etc/sysconfig/chef-server.old`. If there was some `/etc/apache2/vhosts.d/chef-server-cluster.conf` it will be used as-is. Modifications done by the script are scoped by marker comments.

`/opt/dell/bin/chef-server-cluster enable 8` will configure the cluster to use 8 workers.

`/opt/dell/bin/chef-server-cluster enable` can be called safely when cluster is already enabled and will re-configure the cluster appropriately depending on the arguments.

`/opt/dell/bin/chef-server-cluster disable` will disable the cluster and re-start single-worker chef-server. The config files are renamed with `.disabled` extension and services are restarted.


skazi0:
[Evaluate performance of admin node on scale cloud](https://trello.com/c/kQ1eLgYY/122-evaluate-performance-of-admin-node-on-scale-cloud)

skazi0:
@nicolasbock "cluster" might be a bit of an overstatement here. Starting `chef-server` with `-c <number>` results in multiple worker processes being spawned. Apache is used in front of this to hide it from the clients and keep the old port working. This change re-configures `chef-server` service to run with multiple workers and adds apache vhost to handle the load balancing.

nicolasbock:
Ah, thanks! Do we need to configure the ports it is listening to then?

PR #2338, opened 2017-10-27
[4.0] Added script for setting up chef cluster

skazi0:
backport of #2337


skazi0:
[Evaluate performance of admin node on scale cloud](https://trello.com/c/kQ1eLgYY/122-evaluate-performance-of-admin-node-on-scale-cloud)

PR #2339, opened 2017-11-20
Set admin server's platform to suse-12.3 after upgrade to Cloud8

jsuchome:


jsuchome:
Oh, no, this has to go to 4.0

jsuchome:
This is the correct one https://github.com/crowbar/crowbar/pull/2340

PR #2341, opened 2018-01-18
install-chef-suse: Configure epmd.socket correctly

vuntz:
With new erlang and rabbitmq-server packages, rabbitmq now depends on
the system-wide epmd service, which will only listen on IP addresses
defined for epmd.socket.

We need epmd to listen on the admin IP address (the one that resolves
the FQDN), so configure things accordingly.

### DIFF
--- a/scripts/install-chef-suse.sh
+++ b/scripts/install-chef-suse.sh
@@ -665,6 +665,16 @@ if [ ! -f /etc/rabbitmq/rabbitmq.config ] ; then
 EOF
 fi
 
+# Tell system-widde epmd to listen on right address (new rabbitmq-server
+# packages use this epmd service)
+mkdir -p /etc/systemd/system/epmd.socket.d
+cat << EOF > /etc/systemd/system/epmd.socket.d/port.conf
+[Socket]
+ListenStream=$IPv4_addr:4369
+FreeBind=true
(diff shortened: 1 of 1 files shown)


scripts/install-chef-suse.sh
```diff
@@ -665,6 +665,16 @@ if [ ! -f /etc/rabbitmq/rabbitmq.config ] ; then
 EOF
 fi
 
+# Tell system-widde epmd to listen on right address (new rabbitmq-server
+# packages use this epmd service)
+mkdir -p /etc/systemd/system/epmd.socket.d
+cat << EOF > /etc/systemd/system/epmd.socket.d/port.conf
+[Socket]
+ListenStream=$IPv4_addr:4369
+FreeBind=true
```
Itxaka:
```
 This option, when enabled, allows sockets to be bound to a non-local, not configured IP address, and hence allows bindings to a particular IP address before it actually becomes available, making such an explicit dependency to the configured address redundant
```
whats the reason for this? Is the network not available by this time?

vuntz: On reboot, the socket units are started before network.target (and wicked), which means before we get the IP address assigned to any interface. So without this, things fail on reboot.

PR #2344, opened 2018-02-12
sledgehammer: wait for network (bsc#985556)

bmwiedemann:
https://bugzilla.suse.com/show_bug.cgi?id=985556

Sometimes our start-up.sh runs before wicked's dhcp client
brought the network up.
In that case the old ping would return immediately with
'connect: Network is unreachable'
so we need to sleep and retry for a while
because without NFS we cannot really do any node discovery.

### DIFF
--- a/sledgehammer/start-up.sh
+++ b/sledgehammer/start-up.sh
@@ -150,8 +150,13 @@ echo "*.* @@${ADMIN_IP}" >> /etc/rsyslog.conf
 service $RSYSLOGSERVICE restart
 
 # Sometimes at this point network is not up yet, wait for it
-echo "Waiting for admin server ($ADMIN_IP) to be reachable; will wait up to 60 seconds..."
-ping -c 1 -w 60 $ADMIN_IP > /dev/null || {
+n=60
+echo "Waiting for admin server ($ADMIN_IP) to be reachable; will wait up to $n seconds..."
+while (( $n > 0 )) && ! ping -q -c 1 -w 1 $ADMIN_IP > /dev/null ; do
+    sleep 1
+    let n--
+done
+(( $n > 0 )) || {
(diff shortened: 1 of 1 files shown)


sledgehammer/start-up.sh
```diff
@@ -150,8 +150,13 @@ echo "*.* @@${ADMIN_IP}" >> /etc/rsyslog.conf
 service $RSYSLOGSERVICE restart
 
 # Sometimes at this point network is not up yet, wait for it
-echo "Waiting for admin server ($ADMIN_IP) to be reachable; will wait up to 60 seconds..."
-ping -c 1 -w 60 $ADMIN_IP > /dev/null || {
+n=60
+echo "Waiting for admin server ($ADMIN_IP) to be reachable; will wait up to $n seconds..."
+while [[ $n -gt 0 ]] && ! ping -q -c 1 -w 1 $ADMIN_IP > /dev/null ; do
```
nicolasbock: Alternatively (which I recently started preferring because it reads more "normal") `while (( $n > 0 )) &&`.

sledgehammer/start-up.sh
```diff
@@ -150,8 +150,13 @@ echo "*.* @@${ADMIN_IP}" >> /etc/rsyslog.conf
...
+    sleep 1
+    let n--
+done
+[[ $n -gt 0 ]] || {
```
nicolasbock: `(( $n > 0 )) ||`

sledgehammer/start-up.sh
bmwiedemann: updated it. would even work without '$' but I left it in for readability

bmwiedemann:
added this into DC8 sleshammer package to test - can be merged later this week then.

bmwiedemann:
saw a passed SOC8 test today, so it doesnt break discovery.

PR #2346, opened 2018-04-09
[4.0] upgrade: fix epmd during upgrade

Itxaka:
After the package upgrade for epmd it will only listen on
127.0.0.1 which will prevent rabbitmq-server for working, as
it will try to obtain a port on all interfaces by asking epmd
on the appropiate interface but will fail to do so on any interface
other than 127.0.0.1

With this we patch epmd to listen on all interfaces after the upgrade
so it will be ready for when rabbit comes up. The systemd override will
get afterwards be set to listen only in the proper interface as soon as
the crowbar cookbook is run on the admin node.

Note: This is only needed in 7 -> 8 upgrade, should not be needed in the future

### DIFF
--- a/scripts/upgrade_admin_server.sh
+++ b/scripts/upgrade_admin_server.sh
@@ -113,6 +113,20 @@ n.save"
     # Signalize that the upgrade correctly ended
     echo "12.3" >> $UPGRADEDIR/admin-server-upgraded-ok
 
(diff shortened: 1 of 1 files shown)


Itxaka:
currently testing it, seems to work as expected

Itxaka:
Tested and works, ready for merge

PR #2348, opened 2018-05-25
[4.0] upgrade: Make sure schemas are properly migrated after the upgrade

jsuchome:
This used to be one of the tasks of crowbar-init (after it would
set up new database), but because we do not need new database,
we've dropped crowbar-init usage completely.

### DIFF
--- a/scripts/upgrade_admin_server.sh
+++ b/scripts/upgrade_admin_server.sh
@@ -119,6 +119,11 @@ ListenStream=[::]:4369
 FreeBind=true
 EOF
 
+    # Make sure to do schema migration properly after packages were upgraded
+    pushd /opt/dell/crowbar_framework
+    sudo -u crowbar RAILS_ENV=production bin/rake crowbar:schema_migrate_prod
+    popd
+
     # cleanup upgrading indication
     # technically the upgrade is not done yet but it has to be
     # done before the reboot


scripts/upgrade_admin_server.sh
```diff
@@ -119,6 +119,10 @@ ListenStream=[::]:4369
 FreeBind=true
 EOF
 
+    # Make sure to do schema migration properly after packages were upgraded
+    pushd /opt/dell/crowbar_framework
+    sudo -u crowbar RAILS_ENV=production bin/rake crowbar:schema_migrate_prod
```
jhesketh: perhaps we should use an absolute path for bin/rake here?

s-t-e-v-e-n-k:
Looks fine to me, do we need to popd after the migration?

scripts/upgrade_admin_server.sh
jsuchome: Hm, good point. I've used this rake as it has been used from this script before (see https://github.com/crowbar/crowbar/pull/2347) and indeed there's `/opt/dell/crowbar_framework/bin/rake` which we package with crowbar-core package... 
It probably does not make a difference in this call though

jhesketh: Fair enough, but we should popd back :-)

jsuchome: done

PR #2350, opened 2018-05-28
[4.0] upgrade: Lock crowbar-ui before admin upgrade

skazi0:
As crowbar-ui from CloudN will be used to handle N->N+1 upgrade,
we need to lock the package so it's not upgraded with the rest of
admin server.

### DIFF
--- a/scripts/upgrade_admin_server.sh
+++ b/scripts/upgrade_admin_server.sh
@@ -78,6 +78,12 @@ upgrade_admin_server()
 
     trap cleanup INT EXIT
 
+    # Lock crowbar-ui package until upgrade is finished.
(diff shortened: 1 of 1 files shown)


scripts/upgrade_admin_server.sh
```diff
@@ -78,6 +78,9 @@ upgrade_admin_server()
 
     trap cleanup INT EXIT
 
+    # lock crowbar-ui package until upgrade is finished
```
jsuchome: could you please expand the comment? Something like _... so that the package intended for next product upgrade is not installed_ ?

skazi0:
@jsuchome this was not tested at all... imo it would be good to do at least some basic testing to make sure it works as expected.

PR #2355, opened 2018-09-14
set channel_max to unlimited (0) as a workaround

guangyee:
Configure frame_max and channel_max to 0 (unlimited) as a workaround to the
problem where certain older versions of ampq is always sending them as 0
to the server, which causes the server to drop the connection immediately
because by default the max allowed values for channel_max and frame_max are
2047 and 131072 respectively on the server side. See

 https://www.rabbitmq.com/configure.html
 https://github.com/ruby-amqp/amqp/blob/0.6.8/lib/amqp/client.rb#L27

### DIFF
--- a/scripts/install-chef-suse.sh
+++ b/scripts/install-chef-suse.sh
@@ -638,9 +638,28 @@ set_step "pre_sanity_checks"
 
 echo_summary "Starting required services"
 
-# Add frame_max setting as environment variable, because it's just a workaround
+# Configure frame_max and channel_max to 0 (unlimited) as a workaround to the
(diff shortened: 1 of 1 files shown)


jgrassler:
I'm not so sure about this one: our RabbitMQ package ships a configuration file that already sets `channel_max` to `0`. Theoretically that should remain in place (I'll try and find out what causes this setting to flip, there must be something...). Also, adding Rick for a second opinion.

jgrassler:
Ok, turns out we accidently used the wrong `rabbitmq-server` package. If we use the proper one (with a `channel_max` setting of `0` this should not occur anymore.

PR #2365, opened 2019-01-15
upgrade: Process the upgrade of rabbitmq and epmd packages in the right order

jsuchome:


scripts/upgrade_admin_server.sh
```diff
@@ -100,6 +100,17 @@ n.save"
 
     # Upgrade the distribution non-interactively
     zypper --no-color --releasever 12.3 ref -f
+
+    systemctl stop rabbitmq-server
+    systemctl stop couchdb
+    sleep 2
+    epmd -kill
```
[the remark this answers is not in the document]
jsuchome: updated

scripts/upgrade_admin_server.sh
```diff
@@ -100,6 +100,19 @@ n.save"
 
     # Upgrade the distribution non-interactively
     zypper --no-color --releasever 12.3 ref -f
+
+    systemctl stop rabbitmq-server
+    systemctl stop couchdb
+    sleep 5
+    systemctl stop epmd
```
skazi0: I think it's better to do:
```
systemctl stop epmd.socket
systemctl stop epmd.service
```

scripts/upgrade_admin_server.sh
```diff
@@ -100,6 +100,19 @@ n.save"
 
     # Upgrade the distribution non-interactively
     zypper --no-color --releasever 12.3 ref -f
+
+    systemctl stop rabbitmq-server
+    systemctl stop couchdb
```
skazi0: Do we really need that?

jsuchome: I think couchdb could start epmd, that's the reason

skazi0: oh, I didn't know that couchdb is implemented in erlang :)

scripts/upgrade_admin_server.sh
```diff
@@ -100,6 +100,20 @@ n.save"
...
+    sleep 5
+    systemctl stop epmd.socket
+    systemctl stop epmd.service
+    sleep 5
+    killall epmd
```
skazi0: What if epmd doesn't react to SIGTERM? shall we wait and try with `killall -9 epmd`? This seems to be the way systemd does stopping by default.

Itxaka: also, shouldn't we check for epmd state before killing it?

scripts/upgrade_admin_server.sh
```diff
@@ -100,6 +100,20 @@ n.save"
 
     # Upgrade the distribution non-interactively
     zypper --no-color --releasever 12.3 ref -f
+
+    systemctl stop rabbitmq-server
+    systemctl stop couchdb
+    sleep 5
```
Itxaka: why the sleep? I thought systemctl stop would hang until it stops or fails (timeout/failure to stop)

scripts/upgrade_admin_server.sh
```diff
@@ -125,6 +139,8 @@ ListenStream=[::]:4369
 FreeBind=true
 EOF
 
+    systemctl start rabbitmq-server
```
Itxaka: will this autostart couchdb?

skazi0: looking at the rpm scripts, couchdb should be (re)started on package upgrade.

scripts/upgrade_admin_server.sh
skazi0: killall is happy to not kill anything :o)

scripts/upgrade_admin_server.sh
skazi0: it seems to be too late for starting rabbitmq. dist-upgrade above includes crowbar packages which try to upload cookbooks and require rabbit to be running.

scripts/upgrade_admin_server.sh
Gsantomaggio: Added it just for testing.

scripts/upgrade_admin_server.sh
Gsantomaggio: We know ::)
This [patch](https://github.com/crowbar/crowbar/blob/1c9008c6d1704361d6d7f5b75d2e061894b805b9/scripts/upgrade_admin_server.sh#L134) does not work correctly.
I think it is missing a `systemctl daemon-reload`

skazi0: isn't daemon-reload done inside the package scripts? do you know which package changes some unit which doesn't get reloaded?

scripts/upgrade_admin_server.sh
jsuchome: Yeah, this might not be necessary

scripts/upgrade_admin_server.sh
jsuchome: @skazi0 my reason for placing it here was that it is after that `/etc/systemd/system/epmd.socket.d/ports.conf` creation

skazi0: I think we should move both parts above zypper dup. I'm testing it right now. Maybe we need another daemon-reload after ports.conf are created.

skazi0: @jsuchome confirmed: we need another `systemctl daemon-reload` after creating the ports.conf file. So the changes needed for this PR are:
1. move ports.conf creation above zypper dup
2. add `systemctl daemon-reload` after ports.conf are created
3. move `systemctl start rabbitmq-server` above zypper dup but below reload from point 2

Gsantomaggio:
good!

PR #2366, opened 2019-01-17
[5.0] Set admin server's platform to suse-12.4 during upgrade to Cloud8

jsuchome:
This action is still executed on SOC8 side, but only as the last
one, before the admin server reboot.

(cherry picked from commit 817ff9e730d4e2a5474fb73b45bb20915d118fc3)

cherry-pick modified to handle 8-9 upgrade instead of original 7-8

Port of https://github.com/crowbar/crowbar/pull/2340

### DIFF
--- a/scripts/upgrade_admin_server.sh
+++ b/scripts/upgrade_admin_server.sh
@@ -65,7 +65,7 @@ upgrade_admin_server()
         exit 1
     fi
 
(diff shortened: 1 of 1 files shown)


jsuchome:
> please also update master so it doesn't show 12.2 :o)

Actually, this won't be needed. I'm quite sure the upgrade from SOC9 will have to use different code

PR #2367, opened 2019-01-17
[5.0] Process the upgrade of rabbitmq and epmd packages in the right order

jsuchome:
Port of https://github.com/crowbar/crowbar/pull/2365

(Temporary includes https://github.com/crowbar/crowbar/pull/2366)


rhafer:
needs rebase

PR #2368, opened 2019-01-29
doc: Add barclamp development exercises

cmurphy:
A hands-on workshop was given a couple of times covering these
exercises. This commit adds the exercises as documentation so that
others can benefit from it non-interactively.

### DIFF
--- a/doc/barclamp.md
+++ b/doc/barclamp.md
@@ -5,6 +5,8 @@ through all the steps neccessary to create a minimal (without UI elements)
 Barclamp for Crowbar. It will then guide you through all the testing, CI and
 review steps until the point where it gets merged into Crowbar.
 
+See also the [advanced development exercises](barclamp_development_exercises.md).
+
 ## What is a Barclamp?
 
 A barclamp is a plugin for Crowbar that configures a particular service or
--- a/doc/barclamp_development_exercises.md
+++ b/doc/barclamp_development_exercises.md
@@ -1,0 +1,355 @@
+# Barclamp Development Exercises
+
+This guide provides explanation and hands-on exercises to help with
+understanding some of the more useful or unintuitive parts of Crowbar
+development.
+
+The exercises here will have you modify the files on-disk on the Crowbar admin
+node. Your development workflow may involve synchronizing development changes
+from your local workstation to the admin node in some other way, this does not
+cover that.
+
+For the sake of practicing, these exercises will only concern the keystone
+barclamp. This way you only need to deploy up through the keystone barclamp
+which will reduce the amount of time it takes to deploy changes.
+
+## Override Config Files
+
+Crowbar leverages the ``oslo.config`` library to support reading configuration
+from directories under ``/etc/<service>/<service>.conf.d``. The service package
+will install the default config file in ``/etc/<service>/<service>.conf`` and a
+small set of overrides in
+``/etc/<service>/<service>.conf.d/010-<service>.conf``. Crowbar then applies its
+service configuration in ``/etc/<service>/<service>.conf.d/100-<service>.conf``
+which overrides all configuration in the default config file or the package's
+override config file. Because Chef is an traditional-style configuration
+management tool, it attempts to converge a server's configuration by reapplying
+recipes regularly, about every 15 minutes or so. This means that any local
+changes you make to the ``100-<service>.conf`` files that Crowbar manages will
+be eventually overwritten. Using override config files is a useful way of
+working around this, since those files will not be overwritten. Override config
+files are also useful for handling any parameters not yet supported by Crowbar,
+or for temporarily applying a setting without going through a barclamp proposal.
+
+### Exercise
+
+Create a file ``/etc/keystone/keystone.conf.d/200-debug.conf``.
+
+In the file, add the content:
+
+```
+[DEFAULT]
+debug = true
+insecure_debug = true
+```
+
+Restart the keystone service:
+
+```
+# systemctl restart apache2
+```
+
+The ``insecure_debug`` option is now enabled in keystone, which Crowbar doesn't
+support. This option is useful for debugging authentication requests on
+non-production systems. You can see it in action by requesting tokens from
+keystone using invalidate credentials or incorrect scopes.
+
+## Adding a Parameter to a Barclamp
+
+Adding a new parameter involves making changes to several components, but is
+ultimately a matter of copying and pasting similar examples. This exercise will
+add the ``insecure_debug`` parameter to the keystone barclamp. In reality we
+should never support setting this option through Crowbar because it is insecure
+for production use, but it is useful as an exercise.
+
+### Exercise
+
+This is done in three stages.
+
+#### Add the parameter to the data bag
+
+Edit the keystone data bag schema in
+``/opt/dell/chef/data_bags/crowbar/template-keystone.schema``. Add a new
+parameter ``insecure_debug`` under the existing ``debug`` parameter:
+
+```
+{
+  "type": "map", "required": true,
+  "mapping": {
+    "id": { "type": "str", "required": true, "pattern": "/^keystone-|^template-keystone$/" },
+    "description": { "type": "str", "required": true },
+    "attributes": { "type": "map", "required": true,
+      "mapping": {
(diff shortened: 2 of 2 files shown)


doc/barclamp_development_exercises.md
```diff
@@ -0,0 +1,295 @@
...
+#### Add the parameter to the data bag
+
+Edit the keystone data bag schema in
+``/opt/dell/chef/data_bags/crowbar/template-keystone.schema``. Add a new
+parameter ``insecure_debug`` under the existing ``debug`` parameter.
+
+Edit the default data bag in
+``/opt/dell/chef/data_bags/crowbar/template-keystone.json``. Add the new
+parameter there as well. Here you should also increment the schema revision
+number.
```
jsuchome: It would be nice bonus to add snippets of `schema` and `json` files

doc/barclamp_development_exercises.md
```diff
@@ -0,0 +1,295 @@
...
+
+Add a schema migration in ``/opt/dell/chef/data_bags/migrate/keystone/``. A
+``*_noop.rb`` migration can be copied for the function prototypes. The file must
+start with the new schema revision number that you incremented in the default
+data bag. The migration needs to take the new parameter from the template
+attributes hash and add it to the attributes hash. Use another migration that
+adds a parameter as an example.
```
jsuchome: Could you use full name (it's an example after all) together with specific migration number?

cmurphy: Can you give me an example of how you would like it to change? Are you looking for:

> Add a schema migration in ``/opt/dell/chef/data_bags/migrate/keystone/302_add_insecure_debug.rb``. The ``300_noop.rb`` migration can be copied...

My only concern is that the number will quickly become out of date.

cmurphy: Tried to add more specifics, let me know what you think.

doc/barclamp_development_exercises.md
```diff
@@ -0,0 +1,341 @@
...
+data bag and should describe the migration action. For example, if the new
+revision number is 302, the schema migration should go in
+``/opt/dell/chef/data_bags/migrate/keystone/302_add_insecure_debug.rb``. The
+upgrade migration needs to take the new parameter from the template attributes
+hash and add it to the attributes hash, and the downgrade needs to do the
+opposite. It will look something like this:
+
+```
+def upgrade(template_attrs, template_deployment, attrs, deployment)
+  attrs["insecure_debug"] = template_attrs["insecure_debug"]
```
jsuchome: Please add a check if the value is already present - that's especially important for the upgrades, as it could happen that the value that actually already exists is rewritten by the empty one from the template. And even experienced developers make that mistake...

doc/barclamp_development_exercises.md
jsuchome: I like it. Yes, the number will become out of date, but I think it's better this way.

doc/barclamp_development_exercises.md
```diff
@@ -0,0 +1,341 @@
...
+
+Also add an ``only_if`` parameter to the ``execute`` resource so that it will
+only run if that flag is not set. The (naive) idea is that the ``execute``
+resource should run once, then the flag should be set, and then on the next Chef
+run it won't be executed. Moreover, if the resource fails, we would want
+execution to stop so that the problem can be corrected, and then we would want
+the next run to try the mapping population again.
+
+Upload the cookbook and reapply the proposal. Notice the proposal does not fail
+on the invalid commad in the ``execute`` resource. Also check the database on
```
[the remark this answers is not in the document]
cmurphy: Fixed

doc/barclamp.md
```diff
@@ -5,6 +5,8 @@ through all the steps neccessary to create a minimal (without UI elements)
 Barclamp for Crowbar. It will then guide you through all the testing, CI and
 review steps until the point where it gets merged into Crowbar.
 
+See also the [advanced development exercises](barclamp_development_exercises.html).
```
dirkmueller: typo - should link to barclamp_development_exercises.md

doc/barclamp_development_exercises.md
```diff
@@ -0,0 +1,341 @@
...
+recipes regularly, about every 15 minutes or so. This means that any local
+changes you make to the ``100-<service>.conf`` files that Crowbar manages will
+be eventually overwritten. Using override config files is a useful way of
+working around this, since those files will not be overwritten. Override config
+files are also useful for handling any parameters not yet supported by Crowbar,
+or for temporarily applying a setting without going through a barclamp proposal.
+
+### Exercise
+
+Create a file ``/etc/keystone/keystone.conf.d/101-debug.conf``.
```
dirkmueller: technically local modifications should be done with a 2xx prefix, 1xx is reserved for config management (crowbar)

doc/barclamp_development_exercises.md
```diff
@@ -0,0 +1,341 @@
...
+for production use, but it is useful as an exercise.
+
+### Exercise
+
+This is done in three stages.
+
+#### Add the parameter to the data bag
+
+Edit the keystone data bag schema in
+``/opt/dell/chef/data_bags/crowbar/template-keystone.schema``. Add a new
```
dirkmueller: I wonder if we should mention crowbar-core/chef/data_bags/ here instead, as the /opt/dell prefix is implementation defined..

doc/barclamp_development_exercises.md
```diff
@@ -0,0 +1,341 @@
...
+
+Edit the keystone config template in
+``/opt/dell/chef/cookbooks/keystone/templates/default/keystone.conf.erb`` to use
+the new parameter.
+
+Chef does not pick up changes to cookbooks automatically, they need to be
+uploaded to the Chef server:
+
+```
+# knife cookbook upload keystone -o /opt/dell/chef/cookbooks
```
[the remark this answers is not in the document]
dirkmueller: yep, barclamp_install should take care of migrations + chef upload, this separate step should be unnecessary.

doc/barclamp_development_exercises.md
```diff
@@ -0,0 +1,341 @@
...
+the proposal is finished running. If you change the value in the raw view of the
+proposal and reapply, the config file will have changed.
+
+#### Add the parameter to the UI
+
+In many cases this is as far as we go, but for some common operations it is good
+to expose the parameter in the "Custom" view of the crowbar UI.
+
+Edit the locales file at
+``/opt/dell/crowbar_framework/config/locales/keystone/en.yml``. Add the new
```
dirkmueller: path relative to git perhaps?

doc/barclamp_development_exercises.md
```diff
@@ -0,0 +1,341 @@
...
+    node.save
+  end
+  action :nothing
+  subscribes :create, "execute[keystone-manage mapping_populate]", :immediately
+end
+```
+
+Make sure the ``ruby_block`` comes before the closing sync mark. The
+``ruby_block`` is a Chef resource, which means it won't be executed until the
+*converge* phase. Now when you upload the cookbook and apply the proposal, the
```
dirkmueller: .. or to put it differnetly, it actually will get run twice: - during compile phase the "block" is being expanded and the expansion is saved (so input parameters are copied and "frozen at that time" and then during the -convergence phase the list of compiled resources are being executed, potentially triggering from the list of resources additional runs (via subscribe/trigger logic)

doc/barclamp_development_exercises.md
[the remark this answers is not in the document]
cmurphy: It still needs to be a separate step because the exercise goes through running the migration first so that we can see the incremental changes.

doc/barclamp_development_exercises.md
cmurphy: I explained why I'm preferring the package-installed location in the intro

doc/barclamp_development_exercises.md
cmurphy: Rephrased this and expanded on it in the intro for this section

doc/barclamp_development_exercises.md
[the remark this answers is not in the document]
cmurphy: Explained both options

PR #2370, opened 2019-02-07
ipv6: Ensure ip version uniformatiy across networks

matthewoliver:
Ensure all networks are using the same address family and fail
validation if the user is attempting to mix and match IPv4 and
IPv6 networks.

In the early versions of IPv6 support to crowbar, we want to
minimalise the amount of extra testing would be required by
CI/CD. So as an initial step we will only support all IPv4 or
all IPv6 networks. This change inforces that in the
network-json-validator

### DIFF
--- a/scripts/network-json-validator
+++ b/scripts/network-json-validator
@@ -188,6 +188,7 @@ class CrowbarNetwork
   attr_reader :conduit
   attr_reader :vlan
   attr_reader :use_vlan
+  attr_reader :ip_version
 
   def initialize name, json
     @name = name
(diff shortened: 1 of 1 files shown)


scripts/network-json-validator
```diff
@@ -431,16 +434,22 @@ def validate_networks databag
...
       net.validate
       @networks[name] = net
     rescue Exception => e
       raise "Cannot validate definition for network '#{name}': #{e.to_s}"
     end
   end
 
+  if ip_versions.length > 1
+    raise "All networks must either IPv4 or IPv6. A mix is not supported"
+  end
```
nicolasbock: Nice! You could even print out a statistic, sort of like: "You specified 3 IPv4 and 2 IPv6 networks. That's crazy!" :wink:

matthewoliver:
Added the network counts to the error message :)

PR #2377, opened 2019-03-20
crowbar: Relax conditional for aborting installation

nicolasbock:
[SCRD-8330] The `install-chef-suse.sh` script is not only called during initial deployment, but also during a Crowbar restore operation. At restore time, the `crowbar-init.service` might not be running anymore and the current conditional testing for whether the installation can continue is too strict in this case. This change relaxes the conditional by unconditionally running the `reset_crowbar` step, while intercepting a
potential failure.

Signed-off-by: Nicolas Bock <nicolas.bock@suse.de>

### DIFF
--- a/scripts/install-chef-suse.sh
+++ b/scripts/install-chef-suse.sh
@@ -344,19 +344,28 @@ EOF
     exit 1
 fi
 
-if ! (systemctl -q is-active crowbar-init.service && reset_crowbar ); then
-
-    cat <<EOF | pipe_show_and_log
+if ! (reset_crowbar 2> /dev/null); then
+    if systemctl --quiet is-active crowbar-init.service; then
+        cat <<EOF | pipe_show_and_log
 Aborting: Can not initialize Crowbar database
 
-Prior running the installation script, please execute at least
+Prior to running the installation script, please execute at least
+(c.f. Chapter 8 of the Deployment Guide):
 
   systemctl start crowbar-init.service
   crowbarctl database create
 
-If you configured an external database,
-there might be an error connecting.
+If you have configured an external database, there might be an error
+connecting.
 EOF
+    else
+        cat <<EOF | pipe_show_and_log
+Aborting: Cannot initialize Crowbar database
+
+If you have configured an external database, there might be an error
+connecting.
+EOF
+    fi
     exit 1
 fi
 


rktidwell:
I see the backup_restore job is failing. Coincidence, does this just need a "recheck", or is this a real issue?

nicolasbock:
> Looks good to me, just two nits:
> 
> * Can you please add the bug ID (SCRD-8330) to the commit message?
Done.

> * There's a typo in the commit message: `isntall-chef-suse.sh` should be `install-chef-suse.sh`
Fixed.

I will push another iteration soon. Thanks!

nicolasbock:
> I see the backup_restore job is failing. Coincidence, does this just need a "recheck", or is this a real issue?

There is a real issue here. Upon restore, crowbar can't connect to the database. I am looking into it.

nicolasbock:
@rktidwell It's passing now (due to https://github.com/crowbar/crowbar-core/pull/1822). Not sure if this is still interesting then. @dirkmueller , I had originally thought of a restore onto an existing crowbar node that was installed already. Is this a scenario we consider? Or do we only support a restore operation on a newly installed crowbar node (as is tested in CI)?

scripts/install-chef-suse.sh
```diff
@@ -344,18 +344,19 @@ EOF
...
+Prior to running the installation script, please execute at least
+(c.f. Chapter 8 of the Deployment Guide):
 
   systemctl start crowbar-init.service
   crowbarctl database create
 
-If you configured an external database,
-there might be an error connecting.
+If you configured an external database, there might be an error
+connecting.
```
vuntz: After checking the intent of https://github.com/crowbar/crowbar/pull/2369 (which introduced this code and the bug), I would suggest this instead:

```bash
if ! reset_crowbar; then
    if systemctl -q is-active crowbar-init.service; then
        cat <<EOF | pipe_show_and_log
Aborting: Can not initialize Crowbar database

Prior to running the installation script, please execute at least
(c.f. Chapter 8 of the Deployment Guide):

  systemctl start crowbar-init.service
  crowbarctl database create

If you configured an external database, there might be an error
connecting.
EOF
    else
        cat <<EOF | pipe_show_and_log
Aborting: Can not initialize Crowbar database

If you configured an external database, there might be an error
connecting.
EOF
    fi
    exit 1
fi
```

The rationale is that we always want to do `crowbar_reset`; but the error message to provide clear guidance depends on the situation.

nicolasbock: Done.

scripts/install-chef-suse.sh
```diff
@@ -344,20 +344,29 @@ EOF
     exit 1
 fi
 
-if ! (systemctl -q is-active crowbar-init.service && reset_crowbar ); then
-
+if ! (reset_crowbar 2> /dev/null); then
+  if systemctl --quiet is-active crowbar-init.service; then
```
vuntz: Travis is not happy due to indentation here (and on some other lines)

nicolasbock: I pushed a fixed version.

scripts/install-chef-suse.sh
```diff
@@ -344,19 +344,28 @@ EOF
     exit 1
 fi
 
-if ! (systemctl -q is-active crowbar-init.service && reset_crowbar ); then
-
-    cat <<EOF | pipe_show_and_log
+if ! (reset_crowbar 2> /dev/null); then
```
vuntz: Do we really need to hide the output?

jgrassler: Regardless of whether we do or don't need to hide it please come to a consensus. For according to QA this _does_ fix a problem... :-)

nicolasbock: Well, not really. I can unhide it :smile:

nicolasbock:
> No need to block on my comment, but I'd like to have a followup for this...

I pushed #2383 .

PR #2378, opened 2019-04-01
ipv6: Update start-up.sh to be IPv6 compatible

matthewoliver:
The slegehammer/start-up.sh script that is used in the sleshammer
discovery image only supports IPv4 addresses.

This change adds the required IPv6 support. The most notible addition
is how we determine the admin IP address so we can mount the nfs mounts,
etc.
In IPv4 we use wicked's wickedd-dhcp4 binary to get network information
and use the returned SERVERID as the admin ip.
However when using wickedd-dhcp6 the SERVERID ends up just being
something that looks similar to the DUID for the server, NOT the IP.
The only other IP's that could gleaned from this return output is the DNS
address or to grab the pxe boot url and try and grab the host out of that.
But I worry there would be deployments where all these services would be handled
by other servers.

So in order to obtain the admin ip, this patch adds an optional extra
cmdline option 'adminip' which can be added to the kernel commandline.
If there is this script has been modified to use that, otherwise it falls
back to the SERVERID.

I have modified the provisioner barclamp, in crowbar-core, to append the
'adminip=' option to the grub2.cfg and pxelinux.cfg/default files for
the discovery image. And will push the PR as soon as this patch has
landed.

I've also branched the cloud9 sleshammer package in IBS and updated it with
this version of the start-up.sh script, and again will submitpac to staging
as soon as this lands.

I have tested the branched sleshammer image with the updated start-up.sh
file on my local mkcloud instance on my laptop, and it now mounts the nfs
shares successfully in an IPv6 environment.

### DIFF
--- a/sledgehammer/start-up.sh
+++ b/sledgehammer/start-up.sh
@@ -53,10 +53,17 @@ if [[ $(cat /proc/cmdline) =~ $state_re ]] && [ ${BASH_REMATCH[1]} == discovery 
     fi
 fi
 
+# if the admin ip was passed on the cmdline grab it
+admin_ip_re='adminip=([^ ]+)'
+if [[ $(cat /proc/cmdline) =~ $admin_ip_re ]]; then
+    ADMIN_IP="${BASH_REMATCH[1]}"
+fi
+
 # Figure out where we PXE booted from.
 MAC=
 bootif_re='BOOTIF=([^ ]+)'
 ip_re='inet ([0-9.]+)/([0-9]+)'
+ip6_re='inet6 ([0-9a-f:]+)/([0-9]+)'
 if [[ $(cat /proc/cmdline) =~ $bootif_re ]]; then
     MAC="${BASH_REMATCH[1]//-/:}"
     MAC="${MAC#*:}"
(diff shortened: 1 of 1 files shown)


sledgehammer/start-up.sh
```diff
@@ -107,28 +114,43 @@ else
...
+        echo "Things will end badly."
+    fi
 fi
 MYIP="${BASH_REMATCH[1]}"
 
 if suse_ver 12; then
     [ -f /etc/sysconfig/network/config ] && source /etc/sysconfig/network/config
     WAIT_FOR_INTERFACES=${WAIT_FOR_INTERFACES:-120}
     /usr/lib/wicked/bin/wickedd-dhcp4 --test --test-output /tmp/wicked-dhcp-$BOOTDEV --test-timeout $WAIT_FOR_INTERFACES $BOOTDEV
+    if (( $? > 0 )); then
```
s-t-e-v-e-n-k: Anti-pattern.

if ! /usr/lib/wicked/bin/wickedd-dhcp4 --test --test-output /tmp/wicked-dhcp-$BOOTDEV --test-timeout $WAIT_FOR_INTERFACES $BOOTDEV; then

*However* since the command-line arguments so incredibly similar, refactor them into DHCP_ARGUMENTS or such

sledgehammer/start-up.sh
```diff
@@ -107,28 +114,43 @@ else
...
+        /usr/lib/wicked/bin/wickedd-dhcp6 --test --test-output /tmp/wicked-dhcp-$BOOTDEV --test-timeout $WAIT_FOR_INTERFACES $BOOTDEV
+    fi
     source /tmp/wicked-dhcp-$BOOTDEV
-    ADMIN_IP=$SERVERID
+    if [ -z $ADMIN_IP ]; then
```
s-t-e-v-e-n-k: Always quote variables for -z and -n.

sledgehammer/start-up.sh
```diff
@@ -107,28 +114,43 @@ else
     fi
 fi
 
+IPv6=0
 if ! [[ $(ip -4 -o addr show dev $BOOTDEV) =~ $ip_re ]]; then
-    echo "We did not get an address on $BOOTDEV"
-    echo "Things will end badly."
+    if [[ $(ip -6 -o addr show dev $BOOTDEV) =~ $ip6_re ]]; then
+        IPv6=1
+    else
```
s-t-e-v-e-n-k: This feel wrong, somehow. I don't like the nesting either. I wonder if bash has the concept of !~ like Perl does.

But I also can't think of any better way to implement. :-(

s-t-e-v-e-n-k:
Also, the travis failures are real, and you should listen to bashate.

sledgehammer/start-up.sh
matthewoliver: done

sledgehammer/start-up.sh
matthewoliver: done

sledgehammer/start-up.sh
matthewoliver: yeah, I guess we cloud not nest. But we do want to track the ip_version and we get that from here, so could use some else ifs. But it'll end up logically the same

nicolasbock:
Ok, let's do this!

PR #2381, opened 2019-04-09
install-chef-suse: Create /etc/crowbarrc if missing

vuntz:
In case /etc/crowbarrc is missing, the installation would fail. Instead,
simply create it with the default credentials.


jsuchome:
How could it be missing if it is packaged?

vuntz:
> How could it be missing if it is packaged?

It's more for the "I develop from git and without packages" use case (since we still have this working in the script, apparently)

I agree it's not relevant for the product, which is why I didn't think about it in the first place.

PR #2385, opened 2019-04-30
Update barclamp_development_exercises.md

cdevitasuse:
corrected  schema migration path


nicolasbock:
@toabctl Can you merge this? It looks like @cdevitasuse is not a member of the right organizations.

PR #2386, opened 2019-05-27
Separate bmc and admin networks (bsc#1108339)

sandonovsuse:
If bmc and admin networks are on different subnets and different
physcial interfaces, do not enforce bmc VLAN validation.

network-json-validator: remove bmc VLAN validation
install-chef-suse: add bmc VLAN validation if bmc_vlan and admin use the
same interface


JanZerebecki:
It seems nobody would make use of this, as there are no new deployments as the product is discontinued, so closing. If you feel otherwise, please reopen.

The change looks good, it would need to go to master, first.

PR #2387, opened 2019-06-03
IPv6: Export ip_version and handle the DHCP domain

matthewoliver:
The sleshammer image runs the start-up.sh script. This bootstraps
the discovery image on the node and then passes to crowbar-core's
control.sh, via nfs, to finish bootrapping the node.
The control.sh script needs to know the IP verson, start-up.sh knows
what it is but this isn't passed on to control via the environment.

This patch exports the IP_VERSION and ADMIN_IP_WRAPPED variables into
the env so control.sh can pick them up and do the right the righ thing.

The wicked dhcp6 client doesn't generate a DNSDOMAIN variable, so
start-up.sh fails to generate a correct FQDN. In dhcpd6 you can
use the DNS Search domains option which is picked up as DNSSEARCH in wicked.

To support IPv6, if the DNSDOMAIN is missing the script will look for
DNSSEARCH and use the first domain in the list. Note in crowbar we'll only
be passing in one domain. If that fails instead of creating an incorrect
FQDN the full hostname will only contain the short hostname,
ie missing the domain.

### DIFF
--- a/sledgehammer/start-up.sh
+++ b/sledgehammer/start-up.sh
@@ -114,10 +114,10 @@ else
     fi
 fi
 
-ip_version=4
+IP_VERSION=4
 if ! [[ $(ip -4 -o addr show dev $BOOTDEV) =~ $ip_re ]]; then
     if [[ $(ip -6 -o addr show dev $BOOTDEV) =~ $ip6_re ]]; then
-        ip_version=6
+        IP_VERSION=6
     else
         echo "We did not get an address on $BOOTDEV"
         echo "Things will end badly."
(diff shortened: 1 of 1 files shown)


sledgehammer/start-up.sh
```diff
@@ -128,12 +128,16 @@ MYIP="${BASH_REMATCH[1]}"
...
     WAIT_FOR_INTERFACES=${WAIT_FOR_INTERFACES:-120}
-    /usr/lib/wicked/bin/wickedd-dhcp$ip_version --test --test-output /tmp/wicked-dhcp-$BOOTDEV --test-timeout $WAIT_FOR_INTERFACES $BOOTDEV
+    /usr/lib/wicked/bin/wickedd-dhcp$IP_VERSION --test --test-output /tmp/wicked-dhcp-$BOOTDEV --test-timeout $WAIT_FOR_INTERFACES $BOOTDEV
     source /tmp/wicked-dhcp-$BOOTDEV
     if [ -z "$ADMIN_IP" ]; then
         ADMIN_IP=$SERVERID
     fi
-    DOMAIN=$DNSDOMAIN
+    if [ -z "$DNSDOMAIN" ]; then
+        DOMAIN=${DNSSEARCH%%,*}
```
s-t-e-v-e-n-k: I understand what this line does, but I'd very much prefer a short comment to save future confusion.

matthewoliver: Done

